PR #3, opened 2022-06-15
Add JCAD file

trungleduc:


trungleduc:
cc @hbcarlos, I think this PR is ready for review. It's still broken in RTC mode but we need more fixes upstream. Let's merge it first then iterate later

hbcarlos:
Thanks, @trungleduc! Looks good to me!

PR #7, opened 2022-10-04
Use only one shader for all pieces for now

martinRenou:
For now there is only one shader (one material) needed, so no need for compiling as many shader programs as there are meshes.


martinRenou:
Mmh, actually I will close this because I will need one material per-mesh in another coming PR.

PR #8, opened 2022-10-05
Implement highlight on hover using a Color Picking technique

martinRenou:
For now this is only a visual effect, the state is not synced between clients. We can discuss in an issue whether this could be useful or not.

https://user-images.githubusercontent.com/21197331/194020360-5515801c-fee8-4d9d-99ef-f8d40dcdc568.mp4

### DIFF
--- a/src/mainview.tsx
+++ b/src/mainview.tsx
@@ -56,6 +56,8 @@ export class MainView extends React.Component<IProps, IStates> {
       loading: true
     };
 
+    this._pointer = new THREE.Vector2();
+
     this._context = props.context;
     this._cameraClients = {};
     this._context.ready.then(() => {
@@ -173,6 +175,10 @@ export class MainView extends React.Component<IProps, IStates> {
       );
       this._gridHelper.geometry.rotateX(Math.PI / 2);
 
+      // Create the scene for the color picking
+      this._pickingScene = new THREE.Scene();
+      this._pickingTexture = new THREE.WebGLRenderTarget(1, 1);
+
       this._scene.add(this._gridHelper);
       this.addSceneAxe(new THREE.Vector3(1, 0, 0), 0x00ff00);
       this.addSceneAxe(new THREE.Vector3(0, 1, 0), 0xff0000);
@@ -207,6 +213,11 @@ export class MainView extends React.Component<IProps, IStates> {
       this._renderer.setClearColor(0x000000, 0);
       this._renderer.setSize(500, 500, false);
       this.divRef.current.appendChild(this._renderer.domElement); // mount using React ref
+
+      this._renderer.domElement.addEventListener(
+        'pointermove',
+        this._onPointerMove.bind(this)
+      );
 
       const controls = new OrbitControls(
         this._camera,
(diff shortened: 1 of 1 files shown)


trungleduc:
Thanks @martinRenou, It looks great! 
For sharing state with other users, I don't think it is useful but is it possible to highlight an object when we select it from the control panel?

martinRenou:
So actually, maybe instead of an "highlight on hover" behavior, it would make sense to implement a click selection, and when you click on the piece it selects it in the control panel?

trungleduc:
Freecad has both hover and selection events, but I think the selection is more useful.

fc

> That way we can both select from the control panel and select from the 3D view.

That's great!

martinRenou:
Let's do just like FreeCad! The color-picking being super cheap, I think we can afford it.

martinRenou:
I implemented both an hover logic and a click logic for selection:

https://user-images.githubusercontent.com/21197331/194053313-84059174-b6af-4217-af0d-5cc29fb7ba17.mp4

We can bind the control panel UI to the actual 3D selection in a separate PR

src/mainview.tsx
```diff
@@ -250,10 +270,79 @@ export class MainView extends React.Component<IProps, IStates> {
...
+    this._renderer.readRenderTargetPixels(
+      this._pickingTexture,
+      0,
+      0,
+      1,
+      1,
+      pixelBuffer
+    );
+
+    const id = (pixelBuffer[0] << 16) | (pixelBuffer[1] << 8) | pixelBuffer[2];
```
trungleduc, line 301: I'm just curious, what does this line mean?

src/mainview.tsx
```diff
@@ -301,13 +390,26 @@ export class MainView extends React.Component<IProps, IStates> {
...
-    if (old) {
-      this._scene.remove(old);
+    if (this._meshGroup !== null) {
+      this._scene.remove(this._meshGroup);
     }
-    const mainObject = new THREE.Group();
-    mainObject.name = 'shape';
+
+    const pickingColor = new THREE.Color();
+    let i = 1;
```
trungleduc, line 410:
```suggestion
```

src/mainview.tsx
```diff
@@ -301,13 +390,26 @@ export class MainView extends React.Component<IProps, IStates> {
...
+
+    this._meshGroup = new THREE.Group();
     Object.entries(payload).forEach(([objName, data]) => {
```
trungleduc, line 413:
```suggestion
    Object.entries(payload).forEach(([objName, data], idx) => {
```

src/mainview.tsx
```diff
@@ -366,6 +470,12 @@ export class MainView extends React.Component<IProps, IStates> {
       model.castShadow = true;
       model.name = objName;
 
+      // Add the mesh to the color picking scene
+      const pickingMaterial = new THREE.MeshBasicMaterial({
+        color: pickingColor.setHex(i)
```
trungleduc, line 475:
```suggestion
        color: pickingColor.setHex(idx + 1)
```

src/mainview.tsx
```diff
@@ -380,10 +490,12 @@ export class MainView extends React.Component<IProps, IStates> {
 
       model.add(mesh);
 
-      mainObject.add(model);
+      this._meshGroup!.add(model);
+
+      i++;
```
trungleduc, line 495:
```suggestion
```

src/mainview.tsx
trungleduc: I think we can replace this counter with the index of the `forEach` loop.

src/mainview.tsx
trungleduc: Thanks!

trungleduc (approved):

trungleduc:
Thanks!

PR #9, opened 2022-10-05
Use THREE.LineSegments instead of THREE.Line for edges

martinRenou:
This way there is a single buffer, a single shader program, and a single draw call on the GPU for all edges of the same piece


trungleduc:
Thanks!

PR #18, opened 2022-10-06
Use rjsf's JupyterLab theme

### LINKED ISSUE #12: The object properties panel does not handle dark theme
Reference:

Screenshot from 2022-10-06 09-30-09

There might be a way to switch from a light to a dark theme already, similar to what we did for the tree view.

But we can also also extends `react-jsonschema-form` so that it uses Lumino widgets? https://react-jsonschema-form.readthedocs.io/en/latest/advanced-customization/custom-themes that way it would integrate better in the Lab UI. This should be a parallel project/repo though.

martinRenou:
Attempting to fix #12 by using https://github.com/deathbeds/jupyterlab-starters/tree/master/packages/jupyterlab-rjsf

https://user-images.githubusercontent.com/21197331/194346158-703bc648-d814-47ad-a316-9ad4b67602e5.mp4

TODO For this PR:
- [x] Disable the "Ok" button of the dialog if the form is not valid
- [x] Use `jupyterlab-rjsf` for the object properties as well
- [x] ~~Hack around the CSS to make the array widget more pretty~~ this could be achieved using the proper rjsf ui schema properties


trungleduc:
>  Use jupyterlab-rjsf for the object properties as well

The object properties and the toolbar dialog use the same `ObjectPropertiesForm` component, so it should work in both cases.

martinRenou:
> The object properties and the toolbar dialog use the same ObjectPropertiesForm component, so it should work in both cases

Indeed, I wasn't sure how to modify the `ObjectPropertiesForm` at first. But I found a reasonable approach now, I will update my PR.

martinRenou:
This is ready for review :)

https://user-images.githubusercontent.com/21197331/194545323-044b9d40-8e49-42f3-a1bc-4d84f13022eb.mp4

package.json
```diff
@@ -53,14 +53,13 @@
     "watch:labextension": "jupyter labextension watch ."
   },
   "dependencies": {
-    "@fluentui/react": "^8.75.1",
+    "@deathbeds/jupyterlab-rjsf": "^1.1.0",
```
trungleduc, line 56: That org name! :)

src/panelview/formbuilder.tsx
```diff
@@ -14,6 +16,29 @@ interface IProps {
   cancel?: () => void;
 }
 
+export const LuminoSchemaForm = (props: React.PropsWithChildren<any>) => {
```
trungleduc, line 19:
```suggestion
export const LuminoSchemaForm = (props: React.PropsWithChildren<any>): JSX.Element => {
```

trungleduc, line 19: I would credit the original source.

src/panelview/formbuilder.tsx
```diff
@@ -112,40 +139,39 @@ export class ObjectPropertiesForm extends React.Component<IProps, IStates> {
   render(): React.ReactNode {
     if (this.props.schema) {
       const schema = { ...this.props.schema, additionalProperties: true };
+
       const submitRef = React.createRef<HTMLButtonElement>();
+
+      const formSchema = new SchemaForm(schema || {}, {
```
trungleduc, line 145:
```suggestion
      const formSchema = new SchemaForm(schema ?? {}, {
```

style/base.css
```diff
@@ -147,29 +154,6 @@ div.field.field-array > label + span {
...
-  min-width: 80px;
-  height: 32px;
-  background-color: rgb(0, 120, 212);
-  color: rgb(255, 255, 255);
-  user-select: none;
-}
-
 .ms-Grid-col.field {
   margin-bottom: 5px !important;
 }
```
trungleduc, line 159:
```suggestion

```

I think we don't need the `ms-` classes anymore

trungleduc (approved):

trungleduc:
Thanks @martinRenou, it looks great!

PR #20, opened 2022-10-12
Bind mesh selection between tree and 3-D view

martinRenou:
This PR also updates react-tree and improve its styling


martinRenou:
Sure! I will push another commit that starts to add commands and wants to add right click behavior in the tree for delete/hide similar to the file view in lab. Feel free to not continue this work if you think it's not the way to go.

trungleduc:
Finally, I decided to split this PR into 2 parts. I'll update the visual first and the mesh selection sync on a separate PR.

martinRenou:
Closing as done separately

PR #22, opened 2022-10-20
Add `@jupyterlab/shared-models` to the `package.json`

jtpio:
The `@jupyterlab/shared-models` was being referenced here without having its dependency listed in the `package.json`:

https://github.com/QuantStack/jupytercad/blob/f4d7bbf1e4382fa6a110656790f6f2b5e7e95292/src/model.ts#L3


trungleduc:
Thanks @jtpio!

trungleduc (approved):

PR #26, opened 2022-11-07
Handle de-select when selecting a non-visible object

martinRenou:
Selecting a non-visible object in the tree view would not deselect the previously selected object in the 3D view, this PR fixes this use-case.


trungleduc (approved):

trungleduc:
Thanks @martinRenou

PR #29, opened 2022-11-09
Display and sync pointers between clients

martinRenou:
This PR:
- [x] Removes the color-picking logic and implements a BVH picking instead, this has the benefit of being allowed to **retrieve the projected pointer position when picking**
- [x] Displays all the user pointers, using the user color
- [x] Removes the "highlight on hover" feature, as it does not give any more visual information than the pointer now
- [x] Re-implements the camera syncing between users, syncing the entire camera position and orientation information (position, rotation, up, target). This will be useful later when implementing a "follow" feature

Demo: (the user on the right side is hovering the meshes with his mouse, and his pointer is visible to any contributor)

[Screencast from 11-09-2022 04:21:43 PM.webm](https://user-images.githubusercontent.com/21197331/200871010-b2686f6e-d6c9-4be3-b24a-46daa0fcce82.webm)

### DIFF
--- a/src/mainview.tsx
+++ b/src/mainview.tsx
@@ -1,12 +1,22 @@
 import { DocumentRegistry } from '@jupyterlab/docregistry';
+
 import * as React from 'react';
+
+import * as Color from 'd3-color';
+
 import * as THREE from 'three';
 import { OrbitControls } from 'three/examples/jsm/controls/OrbitControls';
 import { Line2 } from 'three/examples/jsm/lines/Line2.js';
 import { LineGeometry } from 'three/examples/jsm/lines/LineGeometry.js';
 import { LineMaterial } from 'three/examples/jsm/lines/LineMaterial.js';
-import { v4 as uuid } from 'uuid';
 
+import {
+  computeBoundsTree,
+  disposeBoundsTree,
+  acceleratedRaycast
+} from 'three-mesh-bvh';
+
+import { v4 as uuid } from 'uuid';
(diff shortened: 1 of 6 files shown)


src/mainview.tsx
```diff
@@ -382,20 +357,39 @@ export class MainView extends React.Component<IProps, IStates> {
   private _onPointerMove(e: MouseEvent) {
     const rect = this._renderer.domElement.getBoundingClientRect();
 
-    this._pointer.x = e.clientX - rect.left;
-    this._pointer.y = e.clientY - rect.top;
+    this._pointer.x = ((e.clientX - rect.left) / rect.width) * 2 - 1;
+    this._pointer.y = -((e.clientY - rect.top) / rect.height) * 2 + 1;
```
trungleduc, line 361: I'm just curious, what does this formula mean?

trungleduc:
> * [ ]  Re-implements the camera syncing between users, syncing the entire camera position and orientation information (position, rotation, up, target). This will be useful later when implementing a "follow" feature
> 

This task is done?

src/mainview.tsx
martinRenou: It scales the DOM coordinates into viewport coordinates (in the range `[-1, 1]`)

trungleduc: Thanks!

trungleduc:
Thanks @martinRenou !

trungleduc (approved):

PR #40, opened 2022-12-01
Custom Open Cascade build

### LINKED ISSUE #34: Custom open-cascade build
Currently the WASM part of JupyterCad fills up user memory and can lead to issues on low memory laptops (my browser or other intensive apps like vscode simply crashes from time to time when using JupyterCad).

We should make a custom open-cascade build to reduce the amount of data shipped to the user.

See https://ocjs.org/docs/getting-started/file-size
Also https://github.com/donalffons/opencascade.js/blob/master/website/ocjs-editor-theme/src/customBuild/customBuild.yml

martinRenou:
Fix #34 

This reduces the load time and the memory needed by the application.

We're going from a 45MB wasm file to a 4.9MB file.


package.json
```diff
@@ -28,14 +28,17 @@
...
   },
   "scripts": {
-    "build": "jlpm build:schema && jlpm run build:lib && jlpm run build:worker && jlpm run build:labextension:dev",
+    "build": "jlpm build:schema && jlpm run build:lib && jlpm run build:copy-opencascade && jlpm run build:worker && jlpm run build:labextension:dev",
     "build:schema": "json2ts -i src/schema -o src/_interface --no-unknownAny --unreachableDefinitions --cwd ./src/schema && cd src/schema && node ../../schema.js",
     "build:prod": "jlpm run clean && jlpm build:schema && jlpm run build:worker:prod && jlpm run build:lib && jlpm run build:labextension",
     "build:labextension": "jupyter labextension build .",
     "build:labextension:dev": "jupyter labextension build --development True .",
     "build:lib": "tsc",
+    "build:copy-opencascade": "copyfiles -f src/worker/opencascade/*.js src/worker/opencascade/*.d.ts src/worker/opencascade/*.wasm lib/worker/opencascade",
```
trungleduc, line 37: We're using the `CopyPlugin` in `extension.webpack.config.js` to copy `wasm` files, maybe we only need to update the config in this file instead of adding a copy command here.

trungleduc:
Thanks a lot @martinRenou!

PR #41, opened 2022-12-02
Fix missing FS and symbol

martinRenou:


trungleduc:
Thanks @martinRenou, is there any flag to minify the `js` and `d.ts` files?

martinRenou:
> is there any flag to minify the js and d.ts files?

I don't know, it should be minified in the end with our Webpack setup for production. Also maybe we want to keep the d.ts file human-readable for development?

trungleduc:
> I don't know, it should be minified in the end with our Webpack setup for production.

 👍  Indeed I forgot our beloved tool `webpack`.

PR #52, opened 2022-12-05
Prevent updating refLength on each remove/hide

### LINKED ISSUE #43: Do not automatically move the camera when updating the objects in the 3D view
Currently, when hiding an object we change the camera position, which can be a bit odd for users

martinRenou:
Fix #43


trungleduc (approved):

trungleduc:
Thanks!

PR #58, opened 2022-12-12
Fix: reset state when unfollowing

martinRenou:
This resolves the issue that the mainview was still borded with the collaboartor color after unfollowing


trungleduc:
Thanks!

PR #61, opened 2022-12-15
Take into account the current user for message displaying

martinRenou:
Show the user icon on the left of the message if it's the current user, on the right if it's another user

[Screencast from 2022-12-15 15-19-03.webm](https://user-images.githubusercontent.com/21197331/207884848-deb77bed-2fe0-4e95-8560-070e6d6a9261.webm)

### DIFF
--- a/src/annotation/view.tsx
+++ b/src/annotation/view.tsx
@@ -50,7 +50,7 @@ export const Annotation = (props: IProps): JSX.Element => {
               <Message
                 user={content.user}
                 message={content.value}
-                index={index}
+                self={model.user?.username === content.user?.username}
               />
             );
           })}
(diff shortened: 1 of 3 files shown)


src/annotation/view.tsx
```diff
@@ -50,7 +52,7 @@ export const Annotation = (props: IProps): JSX.Element => {
               <Message
                 user={content.user}
                 message={content.value}
-                index={index}
```
trungleduc, line 53: We could remove `index` from the map callback.

src/jcadplugin/plugins.ts
```diff
@@ -16,12 +16,15 @@ const activate = (
   tracker: WidgetTracker<IJupyterCadWidget>,
   themeManager: IThemeManager
 ): void => {
+  const { user } = app.serviceManager;
```
trungleduc, line 19: IIRC the `_user` property of `AnnotationModel` is also the current user, can we use this property instead of getting it from here?

martinRenou: Thanks I'll have a look

martinRenou:
I took your comments into account! Thanks for the review

trungleduc (approved):

trungleduc:
Thanks!

PR #68, opened 2022-12-16
Types the API and hides Y data structures

hbcarlos:
~There is a small type issue here:~
https://github.com/QuantStack/jupytercad/blob/7d5af28eea58cda1c2b51d3b218672eec3ca811d/src/toolbar/parttoolbar.tsx#L110

~In the type IJCadObject, `shape` is a `Parts | undefined` while in the schema `shape` is a `string`. See:~

https://github.com/QuantStack/jupytercad/blob/7d5af28eea58cda1c2b51d3b218672eec3ca811d/src/toolbar/parttoolbar.tsx#L21

Edit: Never mind, there is no issue at all.

### DIFF
--- a/src/annotation/model.ts
+++ b/src/annotation/model.ts
@@ -34,7 +34,7 @@ export class AnnotationModel {
   }
 
   getAnnotation(id: string): IAnnotation | undefined {
-    const rawData = this._sharedModel.metadata.get(id);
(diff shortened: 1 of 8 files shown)


trungleduc:
Thanks @hbcarlos !

trungleduc (approved):

PR #72, opened 2022-12-19
Fixes 51

### LINKED ISSUE #51: Delete object is broken
Deleting an object actually deletes another object:

https://user-images.githubusercontent.com/4711805/205687829-e31c75ba-e48a-42d0-a289-dc138189ec75.mp4

hbcarlos:
Fixes #51 

Fixes removing objects and adding objects with duplicated names.

### DIFF
--- a/src/model.ts
+++ b/src/model.ts
@@ -265,13 +265,15 @@ export class JupyterCadDoc
   }
 
   removeObjectByName(name: string): void {
-    let index = -1;
-    this._objects.forEach((obj, idx) => {
-      if (index > -1 || obj.get('name') === name) {
-        index = idx;
+    let index = 0;
+    for (const obj of this._objects) {
+      if (obj.get('name') == name) {
+        break;
       }
-    });
-    if (index > -1) {
+      index++;
+    }
+    
+    if (this._objects.length > index) {
       this.transact(() => {
         this._objects.delete(index);
       });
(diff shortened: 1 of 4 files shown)


src/model.ts
```diff
@@ -256,6 +256,14 @@ export class JupyterCadDoc
     return this._metadataChanged;
   }
 
+  objectExists(name: string): boolean {
+    const obj = this._getObjectAsYMapByName(name);
+    if (obj) {
+      return true;
+    }
+    return false;
```
martinRenou, line 264:
```suggestion
    return Boolean(this._getObjectAsYMapByName(name));
```

src/toolbar/operatortoolbar.tsx
```diff
@@ -111,7 +126,14 @@ export class OperatorToolbarReact extends React.Component<IProps> {
           const model = this.props.toolbarModel.sharedModel;
           if (model) {
             model.transact(() => {
-              model.addObject(objectModel);
+              if (!model.objectExists(objectModel.name)) {
+                model.addObject(objectModel);
```
martinRenou, line 130: Should we make the `objectExists` check under `addObject` instead? So we don't need to add that check code everywhere we use `addObject`?

hbcarlos: `addObject` also checks whether it exists or not. The reason I added `objectExists` is to be able to check before trying to add the element and show an error mesage if it exists.

trungleduc:
Could you rebase it to pick up the CI fix?

trungleduc (approved):

trungleduc:
Thanks @hbcarlos

PR #74, opened 2022-12-20
Annotations panel

martinRenou:
Show all the annotations in the left panel

Screenshot from 2022-12-20 14-16-19

We should probably consider having a similar UI to Google docs for removing annotations:
Screenshot from 2022-12-20 14-19-21


martinRenou:
I marked the PR as ready for review.

Things we could do in separate PRs:
- Add a click behavior in the left panel so that when clicking the annotation it expands/blinks in the 3D view and goes into focus
- Improve the UI for removing annotations -> make a clear distinction between closing/removing

trungleduc:
Thanks @martinRenou, could you update the reference snapshot for the UI test?

martinRenou:
Done! Thanks for the review!

PR #80, opened 2022-12-22
Update build occ command

trungleduc:
Users can now add new symbols to the custom occ build by using this syntax:

```bash
yarn run build:opencascade --add `symbol_to_add` `another_symbol` ...
```

The symbols in the `build.yml` file will be deduplicated and sorted.


martinRenou (approved):
Thanks!

PR #83, opened 2022-12-23
Do not reset the grid and axe when the scene is empty

### LINKED ISSUE #66: Grid size is changed on object visibility switch.
cad-bug

hbcarlos:
Fixes #66 

Should we stop re-setting the camera too? See the screencast:

https://user-images.githubusercontent.com/26092748/209331319-14513f5f-b919-4126-bbd6-bc7cbe0f89a0.mov


martinRenou:
I think there are two cases to consider here when there is only one object left:

- when we **hide** the last object: we **should not** reset any of the camera/grid/axes settings
- when we **remove** the last object: we **should** reset the camera/grid and axes settings

The problem in the current implementation is that you cannot spot in which of those two cases you are under the `_shapeToMesh` method.

One way to fix it would be to change the behavior when hiding an object. I believe hiding an object should not trigger a call to `_shapeToMesh`. Instead, it should simply take the ThreeJS Object3D(s) it refers to and set `object.visible = false` (see [this](https://threejs.org/docs/index.html?q=obj#api/en/core/Object3D.visible)). This way we don't go through this `_shapeToMesh` logic of recreating all the `THREE.BufferGeometry` object (which is far from being ideal because we recreate buffers and re-send them to the GPU while it was actually not needed 😢) and resetting the camera/grid/axes when hiding/showing objects: problem fixed!

trungleduc:
> I think there are two cases to consider here when there is only one object left:
> 
> * when we **hide** the last object: we **should not** reset any of the camera/grid/axes settings
> * when we **remove** the last object: we **should** reset the camera/grid and axes settings
> 
> The problem in the current implementation is that you cannot spot in which of those two cases you are under the `_shapeToMesh` method.
> 

While working with the 2D editor, I realize that we should have a unit system for the length and stick with it. Users need to handle by themself the camera in case of creating something too big. In this case, we can have a static grid/axes.

> One way to fix it would be to change the behavior when hiding an object. I believe hiding an object should not trigger a call to `_shapeToMesh`. Instead, it should simply take the ThreeJS Object3D(s) it refers to and set `object.visible = false` (see [this](https://threejs.org/docs/index.html?q=obj#api/en/core/Object3D.visible)). This way we don't go through this `_shapeToMesh` logic of recreating all the `THREE.BufferGeometry` object (which is far from being ideal because we recreate buffers and re-send them to the GPU while it was actually not needed 😢) and resetting the camera/grid/axes when hiding/showing objects: problem fixed!

I thought about showing and hiding objects by using threejs at the start. But since `visible` is a property of the document model and we need to update it on the FCStd file, I decide to treat it as other cases of modifying the document. So we might need to add some logic to filer out the update of visible property from the shared model change signal.

hbcarlos:
> One way to fix it would be to change the behavior when hiding an object. I believe hiding an object should not trigger a call to _shapeToMesh.

> So we might need to add some logic to filer out the update of visible property from the shared model change signal.

I added the logic to filter the object visibility change and modify the view instead of re-rendering everything.

trungleduc:
Thanks @hbcarlos, could you rebase this PR to pick up the new project structure? We should not commit the output of the `opencascade` build

trungleduc (approved):

martinRenou:
The CI failure might be legit?

hbcarlos:
> The CI failure might be legit?

No, is not. The extension is not activated

trungleduc:
> > The CI failure might be legit?
> 
> No, is not. The extension is not activated

so we need your fix on `ydoc`?

src/mainview.tsx
```diff
@@ -764,6 +772,35 @@ export class MainView extends React.Component<IProps, IStates> {
...
+
+      if (!visible) {
+        this._postMessage({
+          action: WorkerAction.LOAD_FILE,
+          payload: {
+            fileName: this._context.path,
+            content: this._model.getContent()
+          }
+        });
+      }
```
martinRenou, line 800: I'm not sure I understand this part. What is it doing?

trungleduc: I think the variable `visible` should be named `visibleChanged`. 
But I'm wondering what is the behavior of this function if I have both `visible` and other keys in the `change.objectChange` list. We should send the message to the worker to re-render the scene in this case.

martinRenou: I'm not sure it matters much. I'm currently working on the guidata options and having the visible property there. So this method will actually be triggered on an `sharedOptionChange` event, not an object change.

trungleduc: We have both `visible` properties in the document and the GUI document. When we click on the visibility icon in the tree view, which `visible` property we should update? If it's both, then the viewer will get both `sharedOptionChange` and `sharedObjectChange`.

martinRenou: I believe it should be the `visible` property from the GUI data that we should change.

martinRenou: I don't know when we would change the visible property from the object data though. I don't understand why freecad has this separation.

Also, we should probably give the possibility for the user to not sync the guidata state.

trungleduc: > I believe it should be the `visible` property from the GUI data that we should change.

Me too! The script you used can update the visibility of objects in the GUI document?

martinRenou: Yes it can :)

trungleduc: So let's get https://github.com/QuantStack/jupytercad/pull/91 in first then we will adapt this PR?

martinRenou: I'd prefer merging #83 first actually, because I already rebased my work on this locally

trungleduc: ok, I will try to fix the integration tests before merging this PR.

martinRenou:
So maybe this PR should actually remove the grid helper?

trungleduc:
> So maybe this PR should actually remove the grid helper?

I'm in favor of removing the grid too. and for the camera, I think we should only scale it once when users open the document.

src/mainview.tsx
hbcarlos: > But I'm wondering what is the behavior of this function if I have both visible and other keys in the change.objectChange list. We should send the message to the worker to re-render the scene in this case.

When changing the visibility of the object from the tree view the transaction has only one change, the `visible` key. I think is not possible to have an event that changes multiple keys when changing the visibility.

trungleduc:
Thanks @hbcarlos !

PR #84, opened 2023-01-02
Annotations: Change remove button to a close button

### LINKED ISSUE #63: Make annotation deletion more explicit
Currently, you delete an annotation by pressing the small cross handle on the top right corner.

To me, this button looked more like a "close" button than a "delete" button. We should probably have both a close button and a delete button, and make it clear what delete does.

martinRenou:
Fix #63

We don't see the mouse in the screencast, I'm:
- Opening the annotation in the 3D view using the circle handler
- Closing it with the close button (the cross)
- Reopening it
- Removing it with the delete button (the bin)

[Screencast from 2023-01-02 17-50-28.webm](https://user-images.githubusercontent.com/21197331/210259757-cd582570-4f82-4c90-8faf-a945985247ff.webm)


trungleduc:
It works great! Thanks @martinRenou. Since I'm a fan of Windows 😃 , I'm feeling more consistent using the bar as the minimize icon and keeping the cross as the delete icon. What do you think?

icon

martinRenou:
> I'm feeling more consistent using the bar as the minimize icon and keeping the cross as the delete icon. What do you think?

It's indeed more consistent with the cross for removing the 3D objects in the tree view 👍🏽 

I'd be more confident with a dialog asking for delete confirmation if we use this UI. With the previous implementation my first impression was that the cross would just close the annotation and allow me to show it again later, not delete it, a delete confirmation prompt would have helped me.

Did you change the icons locally? Could you push it as a commit on my branch?

trungleduc (approved):

PR #85, opened 2023-01-03
Introduce jupytercad-opencascade separate package

martinRenou:
This package will be published on NPM: https://www.npmjs.com/package/jupytercad-opencascade

This allows us to make JupyterCAD not directly dependent on Docker for the dev build, allowing a Binder build without Docker


trungleduc:
Thanks @martinRenou! Should we make `jupytercad` a monorepo?

martinRenou:
> Thanks @martinRenou! Should we make jupytercad a monorepo?

You mean using lerna? Or is the question more like: "Should we move jupytercad-opencascade outside of jupytercad?"?

trungleduc:
yes, I mean lerna since moving `jupytercad-opencascade` outside of `jupytercad` would make the dev process annoying.

martinRenou:
> since moving jupytercad-opencascade outside of jupytercad would make the dev process annoying

I agree.

I don't really know how to setup lerna, I'll have a look. The only thing I'm worried about is that jupyter-releaser would make as many jupytercad-opencascade releases that we make jupytercad releases, and we probably don't want that.

trungleduc:
> I don't really know how to setup lerna, I'll have a look.

Me neither, maybe @hbcarlos has some clues with `voila-gridstack`?

> The only thing I'm worried about is that jupyter-releaser would make as many jupytercad-opencascade releases that we make jupytercad releases, and we probably don't want that.
I'm not very familiar with `jupyter-releaser`, we are forced to bump all the package versions when we make the release?

martinRenou:
> I'm not very familiar with `jupyter-releaser` 

That makes two of us :P 

@jtpio do you know if we can make a monorepo release with jupyter-releaser but bumping only one of the sub-repo?

hbcarlos:
Hi @martinRenou and @trungleduc. I know how to setup Lerna, I did it in voila-gridstack and Jupyterlab-blockly. I'll do it tomorrow

jtpio:
> do you know if we can make a monorepo release with jupyter-releaser but bumping only one of the sub-repo?

With the releaser it's possible to provide a custom `version-cmd`, for example: https://github.com/jupyterlab/jupyterlab/blob/1d36ce596c48401320e6c33b2a4d17d724db1943/pyproject.toml#LL206-L206C11

The command could include `--scope` filters to only target a specific package when using `lerna` to bump the version.

hbcarlos:
I think making the package private from package.json could work.
I can do it tomorrow, I did it before.

martinRenou:
Thanks for your answer @hbcarlos ! Let's go ahead with this PR then, and do the lerna transition in a separate one

> I think making the package private from package.json could work

I believe we don't want to make it private, because we specifically want it to be published for the Binder build to pull it instead of building it.

> With the releaser it's possible to provide a custom version-cmd, for example: https://github.com/jupyterlab/jupyterlab/blob/1d36ce596c48401320e6c33b2a4d17d724db1943/pyproject.toml#LL206-L206C11

Thanks @jtpio I'll have a look.

martinRenou:
@trungleduc Binder seems to build! (though it takes ages to pull the docker image, I suppose dure to freecad?)

https://mybinder.org/v2/gh/martinRenou/jupytercad/jupytercad-opencascade?urlpath=lab

Do you agree that we could merge this and do the lerna transition separately?

trungleduc:
> Do you agree that we could merge this and do the lerna transition separately?

Yes, I'm testing it locally before merging.

martinRenou:
Binder works! Though it seems it's not started in collaborative mode

Screenshot from 2023-01-03 16-13-30

trungleduc:
Yes, I will fix it later.

jupytercad-opencascade/package.json
```diff
@@ -0,0 +1,41 @@
...
+  ],
+  "homepage": "https://github.com/QuantStack/jupytercad",
+  "bugs": {
+    "url": "https://github.com/QuantStack/jupytercad/issues"
+  },
+  "license": "BSD-3-Clause",
+  "author": {
+    "name": "Trung Le",
+    "email": "leductrungxf@gmail.com"
+  },
```
trungleduc, line 17:
```suggestion
  "author": "QuantStack",
```

trungleduc (approved):

trungleduc:
Thanks a lot @martinRenou !

martinRenou:
Thanks!

PR #87, opened 2023-01-03
Fix binder

trungleduc:
I switched to docker in order to use a base image containing already FreeCAD, it helps reduce the image build time on Binder.


binder/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM trungleduc/jupytercad-env:0.1.1
```
martinRenou, line 1: Should we move this docker image under the jupytercad repo?

trungleduc: You mean the docker file or creating `jupytercad` account on the docker hub?

martinRenou: I mean the Dockerfile that was used for publishing `trungleduc/jupytercad-env` on Dockerhub?

trungleduc: ah ok, where should I put it?

martinRenou: Maybe under a `docker` directory on this repo?

martinRenou: For the account on docker hub, we could probably use the https://hub.docker.com/u/quantstack one

trungleduc: Do you have the right to push images to this account?

martinRenou:
I don't see the JupyterCAD extension working on Binder, though it seems to be enabled:
 
Screenshot from 2023-01-05 10-44-29

trungleduc:
That's weird, it worked till yesterday before changing the base image.

PR #89, opened 2023-01-04
Creates monorepo

hbcarlos:


packages/jupytercad/src/_interface/box.d.ts
```diff
@@ -0,0 +1,41 @@
+/* tslint:disable */
```
martinRenou, line 1: This entire `_interface` directory is not to be committed and should be in the `.gitignore`

trungleduc:
@hbcarlos should we close it?

PR #91, opened 2023-01-06
FCstd color support

martinRenou:
This uses the module [OfflineRenderingUtils](https://github.com/FreeCAD/FreeCAD/blob/20dc9d0a61ec1e9b94ac4088e58581229c2adac3/src/Mod/Arch/OfflineRenderingUtils.py#L33) from freecad in order for it to load the `GuiDocument.xml` file that contains information about color and visibility of 3D objects, that freecad does not usually load if there is no GUI for freecad running.


jupytercad/freecad/loader.py
```diff
@@ -50,14 +96,20 @@ def load(self, base64_content: str) -> None:
...
 
+        # Get objects
         fc_file = fc.app.openDocument(tmp.name)
         for obj in fc_file.Objects:
             self._objects.append(self._fc_to_jcad_obj(obj))
+
+        # Get metadata
         self._metadata = fc_file.Meta
+
+        self._options = _guidata_to_options(OfflineRenderingUtils.getGuiData(tmp.name))
```
trungleduc, line 111: We might need `_options` for other kinds of data. Should we store the GUI data in the `gui` key of the `option` dict?

martinRenou: Sounds good 👍🏽

martinRenou:
TODO still:
- visible state does not seem correct at initialization
- shapeToMesh is being called way too often than it should
- allow for picking colors from the GUI?
- the bounding box should be computed from only visible meshes
- what do we do with the visible state coming from the object (not from guidata)?

martinRenou:
This is ready for review.

The two remaining points from the comment above:
- first one is not related to this PR, it should be fixed separately
- second one could be implemented in a separate PR.

jupytercad/fcstd_ydoc.py
```diff
@@ -7,10 +7,10 @@
 class YFCStd(YBaseDoc):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self._ysource = self._ydoc.get_text('source')
-        self._yobjects = self._ydoc.get_array('objects')
-        self._yoptions = self._ydoc.get_map('options')
-        self._ymeta = self._ydoc.get_map('metadata')
+        self._ysource = self._ydoc.get_text("source")
```
trungleduc, line 10: Can we keep using single quotation marks? I'm using [blue](https://pypi.org/project/blue/) as a drop-in replacement for `black` but with single quotation strings and it works pretty well.

jupytercad/freecad/loader.py
```diff
@@ -12,14 +15,67 @@
...
 
 
+def _guidata_to_options(guidata):
+    """Converts freecad guidata into options that JupyterCad understands"""
+    options = {}
+
+    for obj_name, data in guidata.items():
+        obj_options = {}
+
+        # lol?
```
trungleduc, line 33:
```suggestion
        # TODO: Prevent adding this reserved object name on the frontend side.
```

jupytercad/freecad/loader.py
```diff
@@ -12,14 +15,67 @@
...
+
+
+def _options_to_guidata(options):
+    """Converts JupyterCad options into freecad guidata"""
+    gui_data = {}
+
+    for obj_name, data in options.items():
+        obj_data = {}
+
+        # lol
```
trungleduc, line 56:
```suggestion
```

src/types.ts
```diff
@@ -138,15 +138,16 @@ export interface IJupyterCadDoc extends YDocument<IJupyterCadDocChange> {
   addObjects(value: Array<IJCadObject>): void;
   updateObjectByName(name: string, key: string, value: any): void;
 
-  getOption(key: string): any;
-  setOption(key: string, value: any): void;
+  getOption(key: string): JSONValue;
```
trungleduc, line 141: Since we fixed the `guiData` keyword, I think we can update the schema of `jcad` file to have a more explicit interface than the `JSONValue`

trungleduc: I do not mean the content inside the `guiData`, but the `guiData` keyword now is a part of the `jcad` file specification, I think it's better to have it in the `jcad.json` schema file.

trungleduc:
Galata update snapshots

trungleduc:
Galata please update snapshots

trungleduc (approved):

trungleduc:
Thanks @martinRenou

Issue #79, opened 2022-12-22
Add bot to update galata references


trungleduc:
[jupyterlab/maintainer-tools/.github/actions/update-snapshots@main](https://github.com/jupyterlab/maintainer-tools/blob/main/.github/actions/update-snapshots/action.yml)

PR #94, opened 2023-01-11
Toolbar for configuring the grid and axes

hbcarlos:
Creates a toolbar to configure the axes on each view. By default, the axes are hidden and only added by the user.

https://user-images.githubusercontent.com/26092748/212053504-f1609b35-3a64-458b-a6f8-fe45ef6a5ecb.mov


martinRenou:
I'd personally be in favor of not having a grid in the 3D view **at all**. And instead just show 3 axes like freecad does (see my comment here: https://github.com/QuantStack/jupytercad/pull/83#issuecomment-1377332242).

Also, having the grid settings be set that way (through jlab settings) makes it independent from the opened file, which is odd because different 3D objects would have different orientations and sizes: a grid that makes sense for `electric_razor.fcstd` would not make sense for `plane.fcstd`.

martinRenou:
We could also remove the grid and instead use something simple like:

```typescript
const axesHelper = new THREE.AxesHelper(1000000);
scene.add(axesHelper);
```

trungleduc:
> We could also use something simple like the following instead of trying to create 3D axes that are not bound to the 3D view:
> 
> ```ts
> const axesHelper = new THREE.AxesHelper(1000000);
> scene.add(axesHelper);
> ```

I think I could also remove the old axes in this PR, I can add a new one in the corner of the screen. just like in the freecad interface.

hbcarlos:
In my opinion, having the grid and axe helps to situate the objects when creating a 3D object. We should leave it there so users can add or remove it as they wish.

hbcarlos:
> Also, having the grid settings be set that way (through jlab settings) makes it independent from the opened file, which is odd because different 3D objects would have different orientations and sizes: a grid that makes sense for electric_razor.fcstd would not make sense for plane.fcstd

This PR stores the grid's settings on each model and is not persisted or shared.

martinRenou:
> In my opinion, having the grid and axe helps to situate the objects when creating a 3D object. 

There is no grid in freecad and it works just fine. I'd say we follow freecad as much as we can.

src/schema/axe.json
```diff
@@ -0,0 +1,17 @@
+{
```
trungleduc, line 1: The schema files in this directory are meant to be used to define the structure of the JCAD file. Meanwhile `grid` and `axes` are GUI-related things,  I think we might want to separate these two and put the GUI schemas elsewhere.

src/model.ts
```diff
@@ -30,6 +30,8 @@ export class JupyterCadModel implements IJupyterCadModel {
     this.sharedModel.changed.connect(this._onSharedModelChanged);
     this.sharedModel.awareness.on('change', this._onClientStateChanged);
     this.annotationModel = annotationModel;
+
+    this._view = new Map<string, JSONValue>();
```
trungleduc, line 34: Can we keep `JupyterCadModel` only content related and put the GUI state elsewhere? Maybe we might need another model for the main view to control it from other elements of the interface.

trungleduc: Sorry, I don't follow this one. I just take a look at the [NotebookModel](https://github.com/jupyterlab/jupyterlab/blob/d614e293eb4b90055307828d5e81c9661e1147c9/packages/notebook/src/model.ts#L99) of JupyterLab, and I feel like it's only related to the content and the shared model of the notebook.

hbcarlos:
I removed the grid, removed the axe schema, and moved the view-related properties to the view.

src/toolbar/helpertoolbar.tsx
```diff
@@ -0,0 +1,117 @@
...
+  required: ['Size', 'Visible'],
+  additionalProperties: false,
+  properties: {
+    Size: {
+      type: 'number',
+      description: "Axe's size"
+    },
+    Visible: {
+      type: 'boolean',
+      description: 'Whether the axe is visible or not'
```
trungleduc, line 29:
```suggestion
      description: 'Whether the axes are visible or not'
```

src/toolbar/helpertoolbar.tsx
```diff
@@ -0,0 +1,117 @@
...
+}
+
+const FORM_SCHEMA = {
+  type: 'object',
+  required: ['Size', 'Visible'],
+  additionalProperties: false,
+  properties: {
+    Size: {
+      type: 'number',
+      description: "Axe's size"
```
trungleduc, line 25:
```suggestion
      description: "Size of the axes"
```

src/toolbar/helpertoolbar.tsx
```diff
@@ -0,0 +1,117 @@
...
+  private _createSchema(): IState {
+    const axe: AxeHelper = {
+      size: 5,
+      visible: false
+    };
+    this._panel.setView('axe', axe);
+
+    return {
+      Axe: {
+        title: 'Axe Helper',
```
trungleduc, line 60:
```suggestion
        title: 'Axes Helper',
```

src/toolbar/toolbar.tsx
```diff
@@ -5,13 +5,14 @@ import { ToolbarModel } from './model';
...
 import { UserToolbarReact } from './usertoolbar';
+import { HelpersToolbarReact } from './helpertoolbar';
 
 interface IProps {
   toolbarModel: ToolbarModel;
 }
 
 interface IState {
-  selected: 'PART' | 'OPERATOR';
+  selected: 'PART' | 'OPERATOR' | 'HELPERS';
```
trungleduc, line 15: I would call the new toolbar `DISPLAY` since we will add camera-related buttons to this toolbar.

src/widget.tsx
```diff
@@ -55,9 +66,22 @@ export class JupyterCadPanel extends ReactWidget {
     super.dispose();
   }
 
+  getView(key: string): JSONValue | undefined {
+    return this._view.get(key);
+  }
+
+  setView(key: string, value: JSONValue): void {
```
trungleduc, line 73: Can we enforce the typing of `key` here to `'axes'|...` instead of a generic `string`?

trungleduc:
Thanks @hbcarlos, I'm working on #79 so that we can pass the UI test.

src/widget.tsx
hbcarlos: The idea was to use this API for any helper | display related attributes.

trungleduc: I see, but the downstream needs to put the correct key to activate the related function. So I think it's safer to limit the scope of the `key`. We will enlarge this interface when others helpers/settings come.

trungleduc: Thanks Carlos!

trungleduc (approved):

trungleduc:
I'm working on #96, i will merge this PR after updating the snapshots with the bot

trungleduc:
please update snapshots

hbcarlos:
Is running https://github.com/QuantStack/jupytercad/actions/runs/3903700021/jobs/6668429801
Let's wait

trungleduc:
I need to update the script, the bot runs on every comment added after the magic phrase.

trungleduc:
Galata please update snapshot

trungleduc:
Galata please update snapshots, again

martinRenou:
Thanks!

PR #98, opened 2023-01-16
Add extrusion

trungleduc:


martinRenou:
This needs rebasing now :)

martinRenou:
the `yarn.lock` may need to be updated

trungleduc:
> the `yarn.lock` may need to be updated

Nice catch! I really want to create a bot to update lock files by commenting like the Galata bot 😄

PR #102, opened 2023-01-18
Expose JupyterCad 3d view and APIs in notebook

trungleduc:
## API example:
```python
from jupytercad.notebook import CadDocument
a = CadDocument('examples/cut.FCStd')
box = a.get_object('Box')
box.parameters.Height = 1
box.update_object()
```

https://user-images.githubusercontent.com/4451292/219692474-b4da230e-ecb3-49fc-90ad-97fedfedd860.mp4

## Done
- [x] Add mime renderer for `jcad` and `fcstd` content
- [x] Add APIs to interact with JupyterCad objects.
## Todo
- [ ] Make the `Caddocument` constructor waits for the content coming from the frontend. -> **No solution yet**


trungleduc:
@hbcarlos I'm still have issue with file saving with the latest commit.

jupytercad/notebook/objects/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+    def __init__(__pydantic_self__, parent, **data: Any) -> None:   # noqa
+        super().__init__(**data)
+        __pydantic_self__._caddoc = CadDocument()
+        __pydantic_self__._caddoc.add_object(__pydantic_self__)
+        __pydantic_self__._parent = parent
+
+    def _repr_mimebundle_(self, **kwargs):
+        return self._caddoc.render()
+
+    def update_object(self):
```
martinRenou, line 48: I'm wondering if this should be called automatically whenever the underlying parameters changed.

Instead of:
```python
box.parameters.Height = 1
box.update_object()
```
You would just do:
```python
box.parameters.Height = 1
```

This would be closer to the ipywidgets vision, if that is a goal we'd like to follow.

In ipywidgets, you can also hold changes so that we do not crowd the comm channel with too many messages:
```python
with widget.hold_sync():
    widget.value1 = 1
    widget.value2 = 2
    widget.value3 = 3
```

We could maybe think of having something equivalent here.

src/notebookrenderer/index.ts
```diff
@@ -0,0 +1,90 @@
...
+  JupyterFrontEnd,
+  JupyterFrontEndPlugin
+} from '@jupyterlab/application';
+import { ISessionContext, WidgetTracker } from '@jupyterlab/apputils';
+import { IChangedArgs } from '@jupyterlab/coreutils';
+import { MimeDocument } from '@jupyterlab/docregistry';
+import { INotebookTracker } from '@jupyterlab/notebook';
+import { IRenderMimeRegistry } from '@jupyterlab/rendermime';
+import { IRenderMime } from '@jupyterlab/rendermime-interfaces';
+import { IKernelConnection } from '@jupyterlab/services/lib/kernel/kernel';
```
martinRenou, line 12:
```suggestion
import { IKernelConnection } from '@jupyterlab/services';
```

I recall having dedupe issues when doing this kind of nested imports some time ago. Is `IKernelConnection` not exported by `@jupyterlab/services`?

src/notebookrenderer/widgetManager.ts
```diff
@@ -0,0 +1,111 @@
...
+      const awareness = jcadModel.sharedModel.awareness;
+      const _onUserChanged = (user: User.IManager) => {
+        awareness.setLocalStateField('user', user.identity);
+      };
+      user.ready
+        .then(() => {
+          _onUserChanged(user);
+        })
+        .catch(e => console.error(e));
+      user.userChanged.connect(_onUserChanged, this);
```
martinRenou, line 99: Why is this code here? I don't really understand what it has to do with the rest of the `_handle_comm_open` method.

jupytercad/notebook/objects/__init__.py
trungleduc: I would love to do so, but `pydantic` does not have the field observation (I think @davidbrochart  is working on it). 
I use this approach in `JupyterCad` but we definitely should have something as you describe in the standalone version of `ypywidgets` frontend.

src/notebookrenderer/index.ts
trungleduc: Indeed, this comes from the auto-import feature of VS code, I should have used `Kernel.IKernelConnection` instead.

src/notebookrenderer/widgetManager.ts
trungleduc: Basically, this snippet is done by `WebSocketProvider`, in the case of opening a new widget without providing the file path, we don't initialize the `WebSocketProvider` but `jcadModels` (and then the 3D viewer) still expects that its awareness has the user information.

jupytercad/notebook/objects/__init__.py
davidbrochart: See https://github.com/davidbrochart/ypywidgets/issues/6.

jupytercad/notebook/objects/__init__.py
```diff
@@ -0,0 +1,108 @@
...
+
+OBJECT_FACTORY.register_factory(Parts.Part__Box.value, IBox)
+OBJECT_FACTORY.register_factory(Parts.Part__Cone.value, ICone)
+OBJECT_FACTORY.register_factory(Parts.Part__Cut.value, ICut)
+OBJECT_FACTORY.register_factory(Parts.Part__Cylinder.value, ICylinder)
+OBJECT_FACTORY.register_factory(Parts.Part__Extrusion.value, IExtrusion)
+OBJECT_FACTORY.register_factory(Parts.Part__MultiCommon.value, IIntersection)
+OBJECT_FACTORY.register_factory(Parts.Part__MultiFuse.value, IFuse)
+OBJECT_FACTORY.register_factory(Parts.Part__Sphere.value, ISphere)
+OBJECT_FACTORY.register_factory(Parts.Part__Torus.value, ITorus)
```
martinRenou, line 108: This code may be too complex? It's more or less the same as:

```suggestion
CLASSES = {
    Parts.Part__Box.value: IBox,
    Parts.Part__Cone.value: ICone,
    Parts.Part__Cut.value: ICut,
    Parts.Part__Cylinder.value: ICylinder,
    Parts.Part__Extrusion.value: IExtrusion,
    Parts.Part__MultiCommon.value: IIntersection,
    Parts.Part__MultiFuse.value: IFuse,
    Parts.Part__Sphere.value: ISphere,
    Parts.Part__Torus.value: ITorus,
}

def create_object(
        data: Dict, parent: Optional[CadDocument] = None
) -> Optional[PythonJcadObject]:
    object_type = data.get('shape', None)
    name: str = data.get('name', None)
    if object_type and object_type in CLASSES:
        Model = CLASSES[object_type]
        args = {}
        params = data['parameters']
        for field in Model.__fields__:
            args[field] = params.get(field, None)
        obj_params = Model(**args)
        return PythonJcadObject(
            parent=parent,
            name=name,
            shape=object_type,
            parameters=obj_params,
        )

    return None
```

trungleduc: I want to use a factory pattern to register the shapes at the same place as their definition, the `register_factory` method should be called in the appropriate shape classes. But then shape classes are generated automatically and I didn't find a way to add this `register_factory` call to the generated files, so we ended up with this implementation.

Another advantage of the registering method is that we can prevent removing or overwriting existing factories.

martinRenou (approved):
Thanks!

PR #104, opened 2023-01-23
Improve performance

martinRenou:
I've seen slowness in zooming/panning due to too many client state changes. This PR fixes it by not changing the tree selection if it's not needed.


trungleduc (approved):

trungleduc:
Thanks @martinRenou

PR #105, opened 2023-01-23
Only compute the bounding box of the visible objects for the reflength

martinRenou:


martinRenou:
Please update Galata references

martinRenou:
please update snapshots

trungleduc:
Galata please update snapshots

trungleduc (approved):

PR #107, opened 2023-01-24
Cleanup selection logic and throttle some events

martinRenou:
The main view would always deselect the current selection without checking that it needs to do it, this PR fixes it

### DIFF
--- a/src/mainview.tsx
+++ b/src/mainview.tsx
@@ -651,6 +651,7 @@ export class MainView extends React.Component<IProps, IStates> {
 
   private _selectObject(name: string): void {
     const selected = this._meshGroup?.getObjectByName(name);
+
     if (selected) {
       this._selectedMesh = selected as THREE.Mesh<
         THREE.BufferGeometry,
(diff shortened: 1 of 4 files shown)


src/mainview.tsx
```diff
@@ -69,6 +69,25 @@ interface IPickedResult {
...
+        callback(...args);
+      }, delay);
+    } else {
+      last = now;
+      callback(...args);
+    }
+  } as T;
+}
+
+
```
trungleduc, line 90: I would put it in the `tools.ts` file.

hbcarlos (approved):
Thanks @martinRenou

hbcarlos:
Can you add `jupyter_ydoc==0.2.2` as a dependency [here](https://github.com/martinRenou/jupytercad/blob/7dc74b041e2969b5d76da557de9011a7e757c515/pyproject.toml#L10)?
That might fix the tests.

martinRenou:
Thanks Carlos!

PR #108, opened 2023-01-25
Add launcher buttons

### LINKED ISSUE #100: Create new empty file
JupyterCad can only open existing `jcad` or `FCStd` files. We should allow users to create new empty files and then save them in both supported formats.

hbcarlos:
Fixes #100 

The `FCStd` file doesn't work because as you can see [here](https://github.com/hbcarlos/jupytercad/blob/06fe6acad7629f90c9d7a795e3e6ff5786344748/src/jcadplugin/plugins.ts#L83-L101), we create a new file by calling the save method in the contents service and passing an empty document.

The Freecad format is a [standard zip file containing one or more files in a specific structure](https://wiki.freecadweb.org/File_Format_FCStd), there is no easy way of hardcoding this in the frontend.

Maybe this is a good opportunity to improve the RTC API.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,8 @@
         "@jupyterlab/collaboration": "~4.0.0-alpha.17",
         "@jupyterlab/coreutils": "~6.0.0-alpha.17",
         "@jupyterlab/docregistry": "~4.0.0-alpha.17",
+        "@jupyterlab/filebrowser": "~4.0.0-alpha.17",
+        "@jupyterlab/launcher": "~4.0.0-alpha.17",
         "@jupyterlab/mainmenu": "~4.0.0-alpha.17",
         "@jupyterlab/observables": "~5.0.0-alpha.17",
         "@jupyterlab/services": "~7.0.0-alpha.17",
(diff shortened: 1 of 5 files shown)


trungleduc:
> The Freecad format is a [standard zip file containing one or more files in a specific structure](https://wiki.freecadweb.org/File_Format_FCStd), there is no easy way of hardcoding this in the frontend.
> 

I think creating `jcad` file is enough, we can have a `Save as` button in the toolbar and then export the `FCStd` file from the backend.

trungleduc:
Hi @hbcarlos, what is the status of this PR?

hbcarlos:
We can remove the icon for creating `FCStd` files and merge or wait for a solution to this problem upstream.

trungleduc:
I would go with the `jcad` only launcher for now.

trungleduc:
I just tried from Binder and It does not seem to open correctly the existing files.

https://user-images.githubusercontent.com/4451292/216981340-3ef7f9d4-a436-4e14-b544-749c79ea51b5.mp4

hbcarlos:
> I just tried from Binder and It does not seem to open correctly the existing files.

It should work now.

trungleduc:
Thanks @hbcarlos

trungleduc (approved):

PR #109, opened 2023-01-26
Exploded view

### LINKED ISSUE #38: Add exploded view
It's a cool feature of  `opencascade` web. It would be great if we could add it to `jupytercad`

cad3

martinRenou:
Fix #38 

jupytercad-screenshot

TODO:
- [x] Fix annotations positions in exploded view
- [x] Added UI controls for enabling/disabling the exploded view and changing its factor
- [x] ~~Sync exploded view state with other clients (for the follow mode)~~ won't do
- [x] Fix pointer position by taking the exploded view into account
- [x] Update exploded view when mesh visibility changes
- [x] Update exploded view when mesh removal change?


martinRenou:
The last two commits make the 3D pointers aware of the picked mesh, this allows showing the collaborators pointers at the correct position whether we are in exploded view or not:

Screenshot from 2023-01-27 11-35-56

(right side is picking: left side is showing the right picked position relative to the selected mesh)

trungleduc:
Thanks @martinRenou, it looks great!

PR #116, opened 2023-03-09
Remove `pip install -e .` step in the check release workflow

jtpio:
This might help catch potential issues like in https://github.com/QuantStack/jupytercad/blob/3fc5a729af654d676c9db76fc65a28852892ab90/pyproject.toml#L72.

The "Prep Release" and "Publish Release" workflows of the releaser don't install the package in editable mode anymore: https://github.com/jupyter-server/jupyter_releaser/pull/449

Not doing it in the check release workflow should help stay closer to the actual release process.


jtpio:
The check release fails with:

image

martinRenou (approved):
Thanks!

PR #120, opened 2023-04-03
JLab remix

martinRenou:
This builds upon https://github.com/QuantStack/jupytercad/pull/119

It's still very much in progress

PR #135, opened 2023-04-27
Lab app

trungleduc:
A second try of #120, based on #133 

standalone


martinRenou (approved):
Thanks! Looks great :D

PR #140, opened 2023-05-05
Add top menu and launcher

trungleduc:
jcad2

Closes https://github.com/QuantStack/jupytercad/issues/139


packages/jupytercad-app/src/app/plugins/mainmenu/menuWidget.ts
```diff
@@ -0,0 +1,56 @@
...
+    this._createHelpMenu();
+  }
+
+  private _createHelpMenu(): void {
+    this._commands.addCommand('help-menu:one', {
+      label: 'Item One',
+      execute: () => {
+        console.log('One');
+      }
+    });
```
martinRenou, line 21: Is it to be filled later?

trungleduc: I'll implement these items in this PR

martinRenou: Ok 👍🏽

packages/jupytercad-app/style/base.css
```diff
@@ -1,8 +1,9 @@
...
   min-height: 60px !important;
   padding: 10px !important;
-  justify-content: flex-end;
-  padding-right: 40px !important;
+  justify-content: flex-start;
+  padding-left: 40px !important;
+  padding-right: 30px !important;
   border-bottom: none;
   background: #053b3e;
 }
```
martinRenou, line 9: After this line, there are a couple of theme overwrite here:

```
:root {
  --jp-ui-font-color1: rgba(#4e5c65, 1) !important;
}
:root {
  --jp-inverse-layout-color3: #09777c !important;
}
:root {
  --jp-brand-color1: #09777c !important;
}
```

These don't play well with the dark theme

martinRenou (changes requested):
Some comments on the visuals:

# Hovering style of the `File` menu looks incorrect:

Hover style is not taking the full width:
Screenshot from 2023-05-23 15-05-43

Hover style for the `View` menu looks good
Screenshot from 2023-05-23 15-05-50

# Disable menu items depending on the context

It would be great if the shape/ops buttons were disabled when there is no context opened.

Screenshot from 2023-05-23 15-06-16

# Icons for undo and redo are missing

Screenshot from 2023-05-23 15-06-07

packages/jupytercad-app/style/base.css
trungleduc: I'm thinking should we reuse the dark/light theme of JupyterLab? The best way maybe is to create the dark and light theme for JupyterCad

martinRenou: Probably having custom JupyterCad themes would be the way to go

trungleduc: Indeed, we can do it in another PR

packages/jupytercad-app/src/app/plugins/launcher/index.ts
```diff
@@ -0,0 +1,25 @@
...
+  requires: [ILabShell],
+  autoStart: true,
+  activate: (app: JupyterFrontEnd, labShell: ILabShell): void => {
+    labShell.layoutModified.connect(() => {
+      const els = document.getElementsByClassName('jp-Launcher-sectionTitle');
+      const length = els.length;
+      for (let idx = 0; idx < length; idx++) {
+        const element = els.item(idx);
+        if (element) {
+          element.innerHTML = 'Create new project';
```
martinRenou, line 19:
```suggestion
          element.innerHTML = 'Create New Project';
```

martinRenou:
> Isn't here a way to have a CSS selector that checks for the light/dark body data attribute so that we set those variables correctly depending on this?

If we're able to do it this way, we may not need to implement our own themes for JupyterCAD, and we may support any theme.

trungleduc:
> > Isn't here a way to have a CSS selector that checks for the light/dark body data attribute so that we set those variables correctly depending on this?
> 
> If we're able to do it this way, we may not need to implement our own themes for JupyterCAD, and we may support any theme.

Thank you for pointing it out! The dark theme looks a little bit better now

https://github.com/QuantStack/jupytercad/assets/4451292/77d2226a-cc01-47d2-801d-7e97b30902ff

martinRenou:
Niiice!

martinRenou (approved):
Thanks!!

PR #143, opened 2023-05-15
Update annotations position on resize

### LINKED ISSUE #134: Annotations are not positioned correctly after browser window resize
[Screencast from 2023-04-26 17-34-51.webm](https://user-images.githubusercontent.com/21197331/234627142-2610fcd1-b1bf-48c0-9e58-27c505a51682.webm)

martinRenou:
This PR also improves the window resize performance by not updating the entire react component

Fix #134

### DIFF
--- a/packages/jupytercad-extension/src/mainview.tsx
+++ b/packages/jupytercad-extension/src/mainview.tsx
@@ -88,7 +88,6 @@ export class MainView extends React.Component<IProps, IStates> {
     this._geometry = new THREE.BufferGeometry();
     this._geometry.setDrawRange(0, 3 * 10000);
 
-    this._resizeTimeout = null;
     this.props.view.changed.connect(this._onViewChanged, this);
 
     const lightTheme =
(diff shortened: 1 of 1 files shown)


packages/jupytercad-extension/src/mainview.tsx
```diff
@@ -1065,11 +1064,8 @@ export class MainView extends React.Component<IProps, IStates> {
   };
 
   private _handleWindowResize = (): void => {
-    clearTimeout(this._resizeTimeout);
-    this._resizeTimeout = setTimeout(() => {
-      this.forceUpdate();
-      this._updateAnnotation({ updatePosition: true });
-    }, 500);
```
trungleduc, line 1072: We don't need debouncing anymore?

martinRenou: I suppose the debouncing was because `forceUpdate` was CPU/GPU intensive, but we don't need to call forceUpdate IMO.

The window resize is way smoother with these changes, it's immediate.

trungleduc (approved):

PR #151, opened 2023-05-25
Transform controls

martinRenou:
Applying rotation/translation from the UI is currently very precise but cumbersome. I want a way to apply transformations in a more user-friendly way.

> **Warning**
> Early draft, I should finish my multi-selection PR first.

Screenshot from 2023-05-25 10-53-34

TODO:
- [x] Our 3D meshes all have position [0, 0, 0] instead of taking the centroid of geometries, we should take the centroid of geometries in order for the transform control to work properly
- [ ] Transforming from the UI should update the object in the shared model


martinRenou:
Closing in favor of #529

PR #153, opened 2023-05-25
Fix broken SVG

martinRenou:


martinRenou:
For some reason the SVG would not render anymore in the UI because of the missing newline at the end. It also would not render in the Github diff preview.

martinRenou:
Even weirder is that the visual regression tests did not catch this. It did catch it in https://github.com/QuantStack/jupytercad/pull/150 though.

PR #156, opened 2023-05-26
Update screenshot

trungleduc:
Update readme screenshot


martinRenou:
The new screenshot is not using the latest visual changes 😄

trungleduc:
> The new screenshot is not using the latest visual changes 😄

indeed, I forgot the background change. Coming soon™

martinRenou (approved):
Thanks!

PR #158, opened 2023-05-26
Improve lighting

trungleduc:
## Before 
before

## After

after


trungleduc:
Hey please update snapshots

packages/jupytercad-extension/src/mainview.tsx
```diff
@@ -236,9 +236,9 @@ export class MainView extends React.Component<IProps, IStates> {
 
       this._scene = new THREE.Scene();
 
-      this._scene.add(new THREE.AmbientLight(0xffffff, 0.8)); // soft white light
+      this._scene.add(new THREE.AmbientLight(0xffffff, 0.5)); // soft white light
 
-      const light = new THREE.SpotLight(0xffffff, 0.2);
+      const light = new THREE.PointLight(0xffffff, 1);
       light.castShadow = true;
```
martinRenou, line 242: I think we should disable the shadowmap computing. We're not using it so we could save some perf here.

packages/jupytercad-extension/src/mainview.tsx
```diff
@@ -236,9 +236,9 @@ export class MainView extends React.Component<IProps, IStates> {
 
       this._scene = new THREE.Scene();
 
-      this._scene.add(new THREE.AmbientLight(0xffffff, 0.8)); // soft white light
+      this._scene.add(new THREE.AmbientLight(0xffffff, 0.5)); // soft white light
 
-      const light = new THREE.SpotLight(0xffffff, 0.2);
+      const light = new THREE.PointLight(0xffffff, 1);
```
martinRenou, line 241: I did not know adding a point light to the camera would make the point light stick to the camera position! That's good to know!

martinRenou (approved):
Thanks!

PR #161, opened 2023-05-29
Fixup visibility checking

martinRenou:
The logic should be:
- Set `visible` to True by default
- Check if `object_data.visible` exists, if it does, set `visible` to its value
- Check if `gui_data[object_data['name']].visibilty` exists, if it does, set `visible` to its value

That way, the gui data has precedence over the object data if it exists


martinRenou:
please update snapshots

martinRenou:
https://github.com/QuantStack/jupytercad/actions/runs/5112780603

PR #162, opened 2023-05-29
Python API

martinRenou:
Implementing a simple Python API for shape creation and operators

Screenshot from 2023-05-29 16-15-50

TODO:
- [x] Visual regression tests
- [x] Implement all operators
- [x] Placement should be configurable
- [ ] `update_object` implementation


martinRenou:
please update snapshots

examples/test.jcad
trungleduc, line 1: Why don't we test opening a `jcad` file anymore?

jupytercad/notebook/cad_document.py
```diff
@@ -108,5 +427,102 @@ def _get_yobject_index_by_name(self, name: str) -> int:
...
+            name = f"{obj_type} {n}"
+            n += 1
+
+        return name
+
     def _repr_mimebundle_(self, **kwargs):
         return self.render()
+
+
+class PythonJcadObject(BaseModel):
```
trungleduc, line 445: I'm not sure if we will have the circular import issue but if not, can we split this file into multiple smaller files?

martinRenou: I moved everything in a single file to prevent a circular import indeed.

trungleduc:
The failed integration tests look legit

martinRenou:
Yeah it seems to be since the rebase on master, we now have something printed about the FREECAD path in the Notebook output. We may not want this.

trungleduc:
> Yeah it seems to be since the rebase on master, we now have something printed about the FREECAD path in the Notebook output. We may not want this.

ah ok, it's because we try to import freecad to see if it's available.

trungleduc (approved):

martinRenou:
Sounds good!

PR #163, opened 2023-05-30
Add shadow to annotation boxes

### LINKED ISSUE #157: Annotation has the same background as the 3D viewer
bug-annotation

trungleduc:
Closes #157


packages/jupytercad-extension/style/base.css
```diff
@@ -240,6 +240,8 @@ div.jpcad-toolbar-widget > div.jp-Toolbar-item:last-child {
 .jcad-FloatingAnnotation {
   position: absolute;
   width: 250px;
+  box-shadow: rgba(0, 0, 0, 0.2) 0px 3px 3px -2px,
+    rgba(0, 0, 0, 0.14) 0px 3px 4px 0px, rgba(0, 0, 0, 0.12) 0px 1px 8px 0px;
```
martinRenou, line 244: We should use CSS variables for the shadow. Maybe something like `box-shadow: var(--jp-elevation-z20);` or lower like `--jp-elevation-z12`

martinRenou (changes requested):
Thanks!

martinRenou (approved):
Thanks!

PR #168, opened 2023-05-31
Add backend installation check

### LINKED ISSUE #47: Display nice error message if freecad is not installed
We should show a nice error message to the user if they try to read an FCSTD file wihout freecad installed

trungleduc:
Add an error message for missing FreeCAD installation.

fcmissing

caderror

Closes #47

### DIFF
--- a/jupytercad/notebook/cad_document.py
+++ b/jupytercad/notebook/cad_document.py
@@ -41,17 +41,19 @@ class CadDocument(Widget):
             try:
                 ext = file_name.split(".")[1].lower()
             except Exception:
-                raise Exception("Can not detect file extension!")
+                raise ValueError("Can not detect file extension!")
             if ext == "fcstd":
                 if fc is None:
-                    logger.warn('FreeCAD is required to open FCStd files')
+                    msg = 'FreeCAD is required to open FCStd files'
+                    logger.warn(msg)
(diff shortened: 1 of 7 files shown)


packages/jupytercad-extension/src/fcplugin/plugins.ts
```diff
@@ -55,7 +55,7 @@ const activate = async (
     if (!installed) {
       showErrorMessage(
         'FreeCAD in not installed',
```
martinRenou, line 57:
```suggestion
        'FreeCAD is not installed',
```

martinRenou (commented):
Looks great! Good that we can extend it for other file types later

jupytercad/notebook/cad_document.py
```diff
@@ -41,6 +43,8 @@ def path_to_comm(cls, filePath: Optional[str]) -> Dict:
             except Exception:
                 raise Exception("Can not detect file extension!")
             if ext == "fcstd":
+                if fc is None:
+                    logger.warn('FreeCAD is required to open FCStd files')
```
martinRenou, line 47: We may want to raise an `RuntimeException` here

martinRenou (approved):
Thanks!

PR #169, opened 2023-06-01
Add some documentation

martinRenou:
Would fix https://github.com/QuantStack/jupytercad/issues/149

TODO
- [ ] binder link
- [x] readthedocs bot
- [x] readthedocs config

PR #177, opened 2023-06-02
Add docstrings

martinRenou:
Extracting part of #169 to make a first patch releases with docstrings

### DIFF
--- a/jupytercad/notebook/cad_document.py
+++ b/jupytercad/notebook/cad_document.py
@@ -444,7 +444,7 @@ class CadDocument(Widget):
         rotation_angle: float = 0,
     ) -> CadDocument:
         """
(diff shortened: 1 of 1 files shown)


trungleduc (approved):

jupytercad/notebook/cad_document.py
```diff
@@ -337,11 +443,23 @@ def intersect(
         rotation_axis: List[float] = [0, 0, 1],
         rotation_angle: float = 0,
     ) -> CadDocument:
+        """
+        Apply a intersection boolean operation between two objects. If no objects are provided as input, the last two created objects will be used as operands.
```
trungleduc, line 447:
```suggestion
        Apply an intersection boolean operation between two objects. If no objects are provided as input, the last two created objects will be used as operands.
```

PR #179, opened 2023-06-02
Add Jupyverse plugin

davidbrochart:


davidbrochart:
Yes, we do that in [jupyverse](https://github.com/jupyter-server/jupyverse/blob/ab09a2ea41fcdab226ec55541a26329c3d5b7583/pyproject.toml#L140). Good point, I'll update `pyproject.toml`.

davidbrochart:
This line will have to be added in Jupyter Releaser's `PYPI_TOKEN_MAP` actions secret :
`QuantStack/jupytercad/jupyverse,pypi-token`
Where `pypi-token` is your PyPI token.

trungleduc:
> This line will have to be added in Jupyter Releaser's `PYPI_TOKEN_MAP` actions secret: `QuantStack/jupytercad/jupyverse,pypi-token` Where `pypi-token` is your PyPI token.

@martinRenou could you generate a new PyPI token for the token map secret?

davidbrochart:
Maybe there is a conflict with `hatch-nodejs-version` in the "parent" package (jupytercad) and the dependency (fps-jupytercad) which uses raw `hatch`?

martinRenou:
The bot should now be able to push a release of fps-jupytercad, thanks David!

davidbrochart:
Maybe we should not include `fps-jupytercad` in the Jupyter Releaser for now, and release it manually?
I don't understand the [failing check](https://github.com/QuantStack/jupytercad/actions/runs/5189931476/jobs/9383333026).

pyproject.toml
```diff
@@ -91,6 +91,10 @@ build_dir = "jupytercad/labextension"
 
 [tool.jupyter-releaser.options]
 version-cmd = "python scripts/bump-version.py"
+python_packages = [
+    "jupyverse:fps-jupytercad",
+    ".:jupytercad:fps-jupytercad"
```
trungleduc, line 96: Why do have 2 entries for `fps-jupytercad`?

trungleduc:
> Maybe we should not include `fps-jupytercad` in the Jupyter Releaser for now, and release it manually? I don't understand the [failing check](https://github.com/QuantStack/jupytercad/actions/runs/5189931476/jobs/9383333026).

The failed `check_release` test will also prevent releasing `jupytercad`

pyproject.toml
davidbrochart: First line means that `fps-jupytercad` is in the `jupyverse` directory.
Second line means that `jupytercad` is in the root directory, and depends on `fps-jupytercad`.

martinRenou:
My personal preference is to have fps_jupytercad under the jupytercad repo. If that just means adding a bump script under that directory, I think we can do it.

pyproject.toml
davidbrochart: Yes, and `fps-jupytercad` manages versions through `hatch`, I don't understand why it doesn't work.
I tried https://github.com/QuantStack/jupytercad/pull/179/commits/2957b56717940a398d9df628b30104fcf0e66b11.

trungleduc:
🚀 Thanks @davidbrochart

PR #198, opened 2023-06-10
Add annotation and color API

trungleduc:
This PR allows users to add annotations and modify shape color with Python API

### DIFF
--- a/jupytercad/notebook/cad_document.py
+++ b/jupytercad/notebook/cad_document.py
@@ -57,7 +57,6 @@ class CadDocument(Widget):
         self._metadata = self.ydoc.get_map("metadata")
         self._options = self.ydoc.get_map("options")
 
-
     @property
     def objects(self) -> List[str]:
         """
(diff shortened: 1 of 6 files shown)


jupytercad/notebook/cad_document.py
```diff
@@ -112,6 +118,87 @@ def add_object(self, new_object: "PythonJcadObject") -> CadDocument:
...
+        :param parent: The object which holds the annotation.
+        :param message: The first messages in the annotation.
+        :param position: The position of the annotation.
+        :param user: The user who create this annotation.
+        :return: The id of the annotation if it is created.
+        """
+        new_id = f"annotation_${uuid4()}"
+        parent_obj = self.get_object(parent)
+        if parent_obj is None:
+            logger.error("Parent object not found")
```
martinRenou, line 141: `raise ValueError`?

packages/jupytercad-extension/src/mainview.tsx
```diff
@@ -932,22 +931,34 @@ export class MainView extends React.Component<IProps, IStates> {
 
     if (guidata) {
       for (const objName in guidata) {
+        const obj = this._meshGroup?.getObjectByName(objName) as
+          | BasicMesh
+          | undefined;
+        if (!obj) {
+          return;
```
martinRenou, line 938:
```suggestion
          continue;
```

martinRenou (changes requested):
Thanks!

PR #201, opened 2023-06-14
Remove guidata when deleting an object

### LINKED ISSUE #195: GUIData should be cleaned up when destroying an object
# Issue 

Guidata is not cleaned up when deleting objects. 
To reproduce:

- Create a box "Box 1"
- hide it from the UI
- Delete it
- Create a new box "Box 1" -> It's hidden! Even though it's not the same box anymore

martinRenou:
Fix #195


trungleduc (approved):

trungleduc:
Thanks!

PR #204, opened 2023-06-29
Allow changing camera settings

martinRenou:
Add Camera settings 

For now we can only change the projection, but we could later change the FOV etc

Screenshot from 2023-06-29 10-36-52
Screenshot from 2023-06-29 10-36-25


trungleduc (approved):

trungleduc:
Thanks @martinRenou

trungleduc:
We may need to fix https://github.com/QuantStack/jupytercad/issues/203 to be able to release

martinRenou:
Let's see if that fixes it https://github.com/QuantStack/jupytercad/pull/205

martinRenou:
please update snapshots

PR #205, opened 2023-06-29
Fix check release

martinRenou:


martinRenou:
That works! The visual regression test failing may be sporadic, let's ignore for now

PR #221, opened 2023-11-13
Update OCC build

martinRenou:


packages/jupytercad-opencascade/lib/index.js
```diff
@@ -0,0 +1,9 @@
+import initOpenCascade from 'opencascade.js';
```
martinRenou, line 1: Try to get rid of these as typings are not forwarded properly and not visible in VSCode

martinRenou:
Remaining TODO:
- Fix CI (pydantic issue?)
- Get rid of manual `lib/index.js` and `lib/index.d.ts`
- Get rid of the full WASM build file from dist

martinRenou:
@trungleduc marking as ready for review! But as discussed, let's merge after your refactor PR

trungleduc (approved):

trungleduc:
Thanks @martinRenou!

martinRenou:
Doesn't build anymore since the rebase
I need to push a commit

martinRenou:
Back to the check-release and integration tests failures, let's merge

PR #222, opened 2023-11-15
Pin pydantic (again)

martinRenou:


martinRenou:
The faulty one is still being pulled

martinRenou:
Good! Looks like we only need to pin for the build time (when running datamodel code generator)

trungleduc (approved):

trungleduc:
Thanks!

PR #226, opened 2023-11-21
Fix CI

### LINKED ISSUE #224: Fix releaser and setup a bot account?

trungleduc:
closes #224

## Code change
- Create `jupytercad` meta package
- Update version bump scripts
- Update CI scripts
- Simplify `@jupytercad/opencascade` build script
- Remove Freecad-related test assets.
- Disable `check-python` step in the check release workflow.


trungleduc:
I'll update the UI test snapshot in the follow-up since the updated Galata bot script needs to be on main

examples/fan.stl
martinRenou, line 1: This file is loaded by the Notebook example, we may want to keep it

packages/opencascade/src/jupytercad.opencascade.d.ts
```diff
@@ -0,0 +1 @@
+export default null as any;
```
martinRenou, line 1: What is the purpose of this? Is this for TypeScript to copy over the js file into `lib/`?

ui-tests/tests/notebook.spec.ts-snapshots/dark-Notebook-ipynb-cell-2-linux.png
martinRenou, line 1: Those reference images diff look weird. We should revert all changes in `ui-tests/tests` before merging, and fix the ui-tests in a separate PR

martinRenou (changes requested):

packages/opencascade/src/jupytercad.opencascade.d.ts
trungleduc: This file is there to make tsc happy and compile without buiding occ.js into 'src'. It is not copied to `lib`. The real file in `lib` is provided by webpack in the occ build process.

martinRenou (approved):
Thank you!

PR #230, opened 2023-12-06
Update worker and 3D viewer

### LINKED ISSUE #236: Switching tabs does not update the main view properly
[Screencast from 2023-12-13 16-59-43.webm](https://github.com/jupytercad/jupytercad/assets/21197331/0768ee89-b219-4511-ba2d-5aa9b28dd6fa)

trungleduc:
### Update JCAD format: 

-  new `outputs` field  with the following schema

```json
    "outputs": {
      "type": "object",
      "additionalProperties": false,
      "patternProperties": {
        "^.*$": {
          "type": "string"
        }
      }
```
This field contains the output of the post-processing operator. The 3D viewer uses this field to render the shapes without re-run all post-processing operators at startup,

- New part type `Post::Operator` with the following schema:
```json
{
  "type": "object",
  "description": "Post::Operator",
  "title": "IPostOperator",
  "required": ["Object"],
  "additionalProperties": true,
  "properties": {
    "Object": {
      "type": "string",
      "description": "The name of input object"
    }
  }
}
```
This is a base part for all post-processing operators, the extensions need to extend this schema with the part prefixed with `Post::` to be picked by `JupyterCad`

### New tokens
- `IJCadFormSchemaRegistryToken`: token allowing extension to register new form schema, this helps create parameter form and dialog.
-  `IJCadExternalCommandRegistryToken`: token allowing extension to add buttons to the JupyterCad toolbar

### New worker system
- `IJCadWorkerRegistryToken` token now exposes a default worker using `occ-worker` and allows registering more workers for the post-processing operator.
- An external worker needs to implement the `IJCadWorker` interface.
- On creation, the 3D viewer will register its message handlers with the default worker and all external workers.

### New rendering routine:
- On startup, the 3D viewer will send a `WorkerAction.LOAD_FILE` message to the default worker, the default worker then returns the parsed OCC shapes with the same flag. The viewer will render the OCC shapes and use data from `outputs` field of the shared model to render the post-processing operators.
- On object change, the 3D viewer will send a `WorkerAction.LOAD_FILE` message to the default worker and get back the parsed OCC shapes and also the BREP strings of objects required in the post-processing operator. 
- These BREP strings will be sent to the registered workers with the ` WorkerAction.POSTPROCESS` message. Once the workers finish processing data, they send back to the 3D view the STL string of processed shapes with the ` MainAction.DISPLAY_POST` message.
- The STL strings of processed shapes are saved into the `outputs` field of the shared model.

closes #236


packages/opencascade/src/index.ts
```diff
@@ -1,10 +1,42 @@
...
+        }
+        if (path.endsWith('.worker.js') && !!worker) {
+          return worker;
+        }
+        return path;
+      },
+      ...module
+    }).then(async oc => {
+      for (const lib of libs) {
+        await oc.loadDynamicLibrary(lib, {
```
martinRenou, line 28: Out of curiosity, what is this and why is it needed?

packages/schema/src/schema/jcad.json
```diff
@@ -18,6 +18,15 @@
         }
       },
       "additionalProperties": false
+    },
+    "outputs": {
```
martinRenou, line 22: We may want to provide a way to clean outputs in the future?

packages/opencascade/src/index.ts
trungleduc: I had an issue with building the Salome extension because of missing interfaces in `opencascade.js` so I tried to remove the `opencascade.js` dependency since we only need the init function. 
It turned out the problem was a missing typescript config in the Salome extension, we can roll back to using the function provided by `opencascade.js`

packages/schema/src/schema/jcad.json
trungleduc: indeed, adding it now, the output needs to be deleted when the operator is removed

martinRenou (approved):
Thanks!

PR #231, opened 2023-12-12
Fix missing dev installation

martinRenou:


trungleduc:
Thanks @martinRenou, since you are here, could you pin the prettier version to 3.1.0  in the pre-commit config file? E.g.

```
  - repo: https://github.com/pre-commit/mirrors-prettier
    rev: v3.1.0
    hooks:
      - id: prettier
```

trungleduc:
Thanks!

trungleduc (approved):

PR #232, opened 2023-12-12
STEP files support

martinRenou:


trungleduc:
you can change `"build"` to `"build_dir"` here to fix the release action

https://github.com/jupytercad/jupytercad/blob/f980c3f1196d24154038f672fbee2e86f4bf66ab/python/jupytercad_app/webpack.config.js#L32C7-L32C12

martinRenou:
please update snapshots

packages/schema/src/model.ts
```diff
@@ -300,9 +300,11 @@ export class JupyterCadDoc
   }
 
   get objects(): Array<IJCadObject> {
-    return this._objects.map(
+    const objs = this._objects.map(
       obj => JSONExt.deepCopy(obj.toJSON()) as IJCadObject
     );
+    console.log(objs);
```
trungleduc, line 306:
```suggestion
```

packages/schema/src/model.ts
```diff
@@ -483,6 +485,76 @@ export class JupyterCadDoc
...
+  }
+
+  get objectsChanged(): ISignal<IJupyterCadDoc, IJcadObjectDocChange> {
+    return this._objectChanged;
+  }
+
+  get objects(): Array<IJCadObject> {
+    const source = this._source.toJSON();
+
+    console.log('Source is', this._source, source);
```
trungleduc, line 510:
```suggestion
```

packages/schema/src/model.ts
```diff
@@ -483,6 +485,76 @@ export class JupyterCadDoc
...
+
+    if (!source) {
+      return [];
+    }
+
+    return [
+      {
+        name: 'Step File', // TODO get file name?
```
trungleduc, line 518: Yes, i think the file name is better

packages/schema/src/model.ts
```diff
@@ -483,6 +485,76 @@ export class JupyterCadDoc
...
+  private _sourceObserver = (events: Y.YEvent<any>[]): void => {
+    const changes: Array<{
+      name: string;
+      key: keyof IJCadObject;
+      newValue: IJCadObject;
+    }> = [];
+    events.forEach(event => {
+      event.keys.forEach((change, key) => {
+        changes.push({
+          name: 'Step File',
```
trungleduc, line 542: Same here

packages/schema/src/schema/any.json
```diff
@@ -2,12 +2,16 @@
...
   "additionalProperties": false,
   "properties": {
-    "Shape": {
+    "Content": {
       "type": "string",
-      "description": "The BREP string of the shape"
+      "description": "The string content of the object"
+    },
+    "Type": {
+      "type": "string",
```
trungleduc, line 13: I think an enum of supported type is better, it helps validate the JCAD file and generate the correct interface for `_loadObjectFile` function

packages/occ-worker/src/occapi.ts
```diff
@@ -349,6 +347,40 @@ export function _loadBrep(arg: {
...
+    return shape;
+  } else {
+    console.error('Something in OCCT went wrong trying to read');
+    return undefined;
+  }
+}
+
+export function _loadObjectFile(arg: {
+  content: string;
+  type: string;
```
trungleduc, line 372: a more specific interface in form of `step | brep` would be better. I goes with the comment in the `any.json` schema.

python/jupytercad_core/src/stepplugin/modelfactory.ts
```diff
@@ -0,0 +1,85 @@
+import { IJupyterCadDoc, JupyterCadModel } from '@jupytercad/schema';
+import { DocumentRegistry } from '@jupyterlab/docregistry';
+import { Contents } from '@jupyterlab/services';
+
+/**
+ * A Model factory to create new instances of JupyterCadModel.
+ */
+export class JupyterCadJcadStepModelFactory
```
trungleduc, line 8:
```suggestion
export class JupyterCadStepModelFactory
```

trungleduc (changes requested):

packages/schema/src/model.ts
martinRenou: Unfortunately the ydoc does not known anything about the name of the file it originated from, or am I missing something?

trungleduc: ah indeed. So let's keep it like this.

trungleduc:
Thanks @martinRenou!

PR #233, opened 2023-12-12
Galata bot: Missing hub command

martinRenou:


martinRenou:
Merging right away as the CI is not testing this

PR #235, opened 2023-12-13
Non-editable step file

martinRenou:


packages/schema/src/interfaces.ts
```diff
@@ -78,6 +78,7 @@ export interface IJupyterCadDoc extends YDocument<IJupyterCadDocChange> {
   objects: Array<IJCadObject>;
   options: JSONObject;
   metadata: JSONObject;
+  editable: boolean;
```
trungleduc, line 81:
```suggestion
  readonly editable: boolean;
```
Should it be read-only?

trungleduc:
Thanks!

PR #238, opened 2023-12-19
Step file export to .jcad

martinRenou:


martinRenou:
It seems I cannot open the resulting exported file and getting the dreaded:

```
if event == "metadata" and self._last_modified < args["last_modified"]:
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    TypeError: '<' not supported between instances of 'NoneType' and 'datetime.datetime'
```

which [should have been fixed in jupyter_collaboration 2.0.0](https://github.com/jupyterlab/jupyter-collaboration/issues/162)

trungleduc:
Everytime I have this error message, the real error is from my code and not from `jupyter_collaboration`

martinRenou:
> Everytime I have this error message, the real error is from my code and not from jupyter_collaboration

It seems that the latest jupyter-collaboration 1.x patch does not hide the real error anymore with this last_modified thingy.

martinRenou:
please update snapshots

martinRenou:
please update snapshots

martinRenou:
https://github.com/jupytercad/jupytercad/actions/runs/7842797150

PR #245, opened 2024-01-04
Don't show properties panel for readonly files

### LINKED ISSUE #240: Error in property panel of a STEP file
We may need to deactivate the property panel of non-editable files

step

martinRenou:
Fix  #240


trungleduc (approved):

trungleduc:
Thanks @martinRenou

PR #333, opened 2024-03-11
Implement clipping

martinRenou:
Remaining TODOs:
- [x] Have a nicer icon
- [x] Have keyboard shortcuts for switching from rotation mode to translate mode
- [x] Use stencil buffer to fill the shapes?

[Screencast from 2024-03-12 14-12-49.webm](https://github.com/jupytercad/jupytercad/assets/21197331/cc8c9817-f021-406a-a48f-a0fcd5e68883)

### DIFF
--- a/packages/base/src/mainview.tsx
+++ b/packages/base/src/mainview.tsx
@@ -417,12 +417,12 @@ export class MainView extends React.Component<IProps, IStates> {
   startAnimationLoop = (): void => {
     this._requestID = window.requestAnimationFrame(this.startAnimationLoop);
 
-    if (this._clipPlane !== null) {
-      this._clippingPlane.coplanarPoint(this._clipPlane.position);
-      this._clipPlane.lookAt(
-        this._clipPlane.position.x - this._clippingPlane.normal.x,
-        this._clipPlane.position.y - this._clippingPlane.normal.y,
-        this._clipPlane.position.z - this._clippingPlane.normal.z
+    if (this._clippingPlaneMesh !== null) {
+      this._clippingPlane.coplanarPoint(this._clippingPlaneMesh.position);
+      this._clippingPlaneMesh.lookAt(
+        this._clippingPlaneMesh.position.x - this._clippingPlane.normal.x,
+        this._clippingPlaneMesh.position.y - this._clippingPlane.normal.y,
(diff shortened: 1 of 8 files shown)


trungleduc:
The `mainview.tsx` file has become quite large, should we extract the ThreeJS-only logic into a separate file/module?

martinRenou:
> should we extract all ThreeJS-only logic

It seems to me the mainview.tsx already contains only ThreeJS specific logic, what other logic do you see could be extracted from this file?

The rendering logic is indeed becoming a bit complex. This PR breaks some selection logic for now, I need to fix it. And adding more rendering features will eventually become tricky.

trungleduc:
I mean we can do a refactoring pass in another PR to move different ThreeJS bits into standalone functions and use them in the main view. At least it will make the mainview.tsx easier to read

trungleduc:
Thanks @martinRenou , it looks great! I wonder if it makes sense to show the clipping plane in transparent?

martinRenou:
> I wonder if it makes sense to show the clipping plane in transparent?

To see "inside" of the mesh?

packages/base/src/types.ts
```diff
@@ -28,6 +28,13 @@ export type CameraSettings = {
   type: 'Perspective' | 'Orthographic';
 };
 
+/**
+ * The state of the clip view
+ */
+export type ClipSettings = {
+  enabled: boolean;
```
trungleduc, line 35: Should we add a setting to show/hide the clipping plane?

packages/base/src/mainview.tsx
```diff
@@ -572,6 +653,25 @@ export class MainView extends React.Component<IProps, IStates> {
       this._model.syncSelectedObject(names, this.state.id);
     }
   }
+
+  private _onKeyDown(event: KeyboardEvent) {
+    // TODO Make these Lumino commands? Or not?
+    if (this._clipSettings.enabled) {
+      switch (event.key) {
+        case 'r':
```
trungleduc, line 661: Could you update the documentation about the clipping feature and the hotkey?

martinRenou: 👍🏽 Shall we do this in a separate PR?

trungleduc (approved):

trungleduc:
Thanks @martinRenou

Issue #343, opened 2024-03-29
Implement proper operator error handling 


martinRenou:
#340 introduced new operators that may fail to compute eventually. 

We should implement proper operator error handling and reporting, so that users understand what happens upon failure.

martinRenou:
cc. @trungleduc 

Following an internal discussion with @SylvainCorlay.

My PR https://github.com/jupytercad/jupytercad/pull/349 to handle error messages from Open Cascade is not a good approach as we do in order:
1. Let the user confirm the operator form
2. Apply the transaction to the shared model
3. React to the shared model change by sending the operation to the OCC worker
4. (New in the PR) show an error dialog upon failure
I believe this is not a good approach because the shared model has been changed with a broken operator. This means the error dialog will show up a lot for the same error (when refreshing, when updating an operand etc).

### Proposed solution

Another way to tackle this Open Cascade failure would be to make a dry run in the OCC worker to try to apply the operator before either step **1.** (prevent form confirmation) or **2.** (prevent shared model change).

Upon success, our worker is already smart enough to not compute the operator a second time by caching the operator output during the dry run, so performance wide this solution is good.

Right now the 3D view owns the OCC worker, AFAIK the operators commands don't have a pointer to it. So there may be some refactor do it here to make it right.

trungleduc:
> Right now the 3D view owns the OCC worker, AFAIK the operators commands don't have a pointer to it. So there may be some refactor do it here to make it right.

No, the OCC worker is accessible anywhere via the `IJCadWorkerRegistry` token

martinRenou:
Nice. I'll have a look at implementing this approach.

Issue #200, opened 2023-06-13
Have a UUID objects


martinRenou:
Renaming an existing object is problematic today if the object is used in a boolean operator.

We should have a UUID attached to each object that is **not** controlled by the user, and boolean operators would use this and not the object names as input.

martinRenou:
As discussed with @trungleduc and @arjxn-py:
- [ ] We might need to start introducing the schema version into the metadata field of jcad file
- [ ] We need to switch the underlying objects structure from a list to a dictionary
- [ ] we could support multiple schema versions at once

trungleduc:
The schema version check should be in `3.0.0`, so that in the future when users open a jcad file with a schema version greater than the supported one, they can get the error dialog.

PR #349, opened 2024-04-16
Error handling

martinRenou:
Fix https://github.com/jupytercad/jupytercad/issues/343

https://github.com/jupytercad/jupytercad/assets/21197331/15a22bb3-46f9-40ca-a960-815b48069a93

### DIFF
--- a/packages/base/src/panelview/objectproperties.tsx
+++ b/packages/base/src/panelview/objectproperties.tsx
@@ -109,9 +109,9 @@ class ObjectPropertiesReact extends React.Component<IProps, IStates> {
       return;
     }
 
-    const currentViewModel = this.props.tracker
+    const currentWidget = this.props.tracker
       .currentWidget as JupyterCadWidget | null;
-    if (!currentViewModel) {
+    if (!currentWidget) {
       return;
     }
 
@@ -133,7 +133,7 @@ class ObjectPropertiesReact extends React.Component<IProps, IStates> {
 
     // Try a dry run
     const dryRunResult =
-      await currentViewModel.content.currentViewModel.dryRun(updatedContent);
+      await currentWidget.content.currentViewModel.dryRun(updatedContent);
     if (dryRunResult.status === 'error') {
       showErrorMessage(
         'Failed to update the shape',
@@ -142,7 +142,7 @@ class ObjectPropertiesReact extends React.Component<IProps, IStates> {
       return;
     }
 
-    // Dry run was successful, really apply the update now
+    // Dry run was successful, ready to apply the update now
     const obj = model.sharedModel.getObjectByName(objectName);
     if (obj) {
       model.sharedModel.updateObjectByName(objectName, 'parameters', {
--- a/packages/base/src/panelview/objectproperties.tsx
+++ b/packages/base/src/panelview/objectproperties.tsx
@@ -5,9 +5,10 @@ import {
   IJcadObjectDocChange,
   IJupyterCadClientState,
   IJupyterCadDoc,
-  IJupyterCadModel
+  IJupyterCadModel,
+  IJupyterCadTracker
 } from '@jupytercad/schema';
-import { ReactWidget } from '@jupyterlab/apputils';
+import { ReactWidget, showErrorMessage } from '@jupyterlab/apputils';
 import { PanelWithToolbar } from '@jupyterlab/ui-components';
 import { Panel } from '@lumino/widgets';
 import * as React from 'react';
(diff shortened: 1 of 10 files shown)


trungleduc:
Thanks @martinRenou for working on it. I just tested with some corner cases:

- Adding shape or editing shape from the right panel does not go through the dry-run check, should we do it?

https://github.com/jupytercad/jupytercad/assets/4451292/f7d51cba-98f0-4889-b003-2f8af6804189

- I artificially make the `loadFlle` action much slower so that I can add another operator while the first one is still processing. But then the first operator is discarded, we may need to lock the operators while the worker is busy

https://github.com/jupytercad/jupytercad/assets/4451292/e93699ee-5288-436c-8bb3-c6788062bbd3

- When there are 2 users, the first user adds an operator but since the shared doc is not updated, the second user might add another operator with the same name

https://github.com/jupytercad/jupytercad/assets/4451292/17cb5797-a1a6-435c-b262-962d2968b846

- Even if the second user can avoid the naming collision, the outcome is not correct since their worker does not know about the upcoming operator of the first user


https://github.com/jupytercad/jupytercad/assets/4451292/f99fe843-25ca-46dc-96e5-890ff52289b9

I think we might need to improve the loading indicator and locking mechanism to avoid these cases

martinRenou:
> Adding shape or editing shape from the right panel does not go through the dry-run check, should we do it?

Yes, definitely. Also we should probably do it on the first load of the file, and report malformed a file if that's the case.

> I artificially make the loadFlle action much slower so that I can add another operator while the first one is still processing. But then the first operator is discarded, we may need to lock the operators while the worker is busy

Right! We should fix it indeed. It's unrelated to this PR though, so we should probably turn this into an issue and look into implementing a fix.

> When there are 2 users, the first user adds an operator but since the shared doc is not updated, the second user might add another operator with the same name

I believe name conflicts should not be a thing. We can fix it by fixing https://github.com/jupytercad/jupytercad/issues/200. 

> Even if the second user can avoid the naming collision, the outcome is not correct since their worker does not know about the upcoming operator of the first user

I guess this would be fixed by implementing a proper lock?

trungleduc:
yeah, having a proper operator lock definitely helps. But I also feel that having a single promise delegate for all operators might be a bottleneck. Using a dict of promise delegates and sending the delegate id back and forth between the worker and the main view model can work?

martinRenou:
Oh I see. I'll have a look at that.

trungleduc (approved):
Thanks 🚀

martinRenou:
Wait before merging I'll check to use the dry run in the object properties panel too :)

martinRenou:
@trungleduc Ready for review!

packages/base/src/panelview/objectproperties.tsx
```diff
@@ -97,18 +101,51 @@ class ObjectPropertiesReact extends React.Component<IProps, IStates> {
...
   ) {
     if (!this.state.jcadObject || !objectName) {
       return;
     }
 
-    const model = this.props.cpModel.jcadModel?.sharedModel;
-    const obj = model?.getObjectByName(objectName);
-    if (model && obj) {
-      model.updateObjectByName(objectName, 'parameters', {
+    const currentViewModel = this.props.tracker
```
trungleduc, line 112:
```suggestion
    const currentWidget= this.props.tracker
```

packages/base/src/panelview/objectproperties.tsx
```diff
@@ -97,18 +101,51 @@ class ObjectPropertiesReact extends React.Component<IProps, IStates> {
...
+      .currentWidget as JupyterCadWidget | null;
+    if (!currentViewModel) {
```
trungleduc, line 114:
```suggestion
    if (!currentWidget) {
```

packages/base/src/panelview/objectproperties.tsx
```diff
@@ -97,18 +101,51 @@ class ObjectPropertiesReact extends React.Component<IProps, IStates> {
...
+        object.parameters = {
+          ...object.parameters,
+          ...properties
+        };
+      }
+    }
+
+    // Try a dry run
+    const dryRunResult =
+      await currentViewModel.content.currentViewModel.dryRun(updatedContent);
```
trungleduc, line 136:
```suggestion
      await currentWidget.content.currentViewModel.dryRun(updatedContent);
```

packages/base/src/panelview/objectproperties.tsx
```diff
@@ -97,18 +101,51 @@ class ObjectPropertiesReact extends React.Component<IProps, IStates> {
...
+    if (dryRunResult.status === 'error') {
+      showErrorMessage(
+        'Failed to update the shape',
+        'The tool was unable to update the desired shape due to invalid parameter values. The values you entered may not be compatible with the dimensions of your piece.'
+      );
+      return;
+    }
+
+    // Dry run was successful, really apply the update now
```
trungleduc, line 145:
```suggestion
    // Dry run was successful, ready to apply the update now
```

SylvainCorlay:
🎉  that is pretty great!

PR #351, opened 2024-04-19
Add await for postShape before sending

gjmooney:
Wait for the `GLTFExporter` callback to populate `postShape` before sending `postResult` to worker threads.

### DIFF
--- a/packages/base/src/3dview/mainview.tsx
+++ b/packages/base/src/3dview/mainview.tsx
@@ -34,8 +34,6 @@ import {
 import { FollowIndicator } from './followindicator';
 import {
   BasicMesh,
-  buildShape,
-  computeExplodedState,
   DEFAULT_EDGE_COLOR,
   DEFAULT_EDGE_COLOR_CSS,
   DEFAULT_LINEWIDTH,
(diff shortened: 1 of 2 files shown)


martinRenou (approved):
LGTM! Thanks

packages/base/src/3dview/mainview.tsx
```diff
@@ -752,10 +758,13 @@ export class MainView extends React.Component<IProps, IStates> {
             threeShape,
             exported => {
               pos.postShape = exported as any;
+              resolved();
```
trungleduc, line 761: it should resolve the promise at the end of this loop?

trungleduc (changes requested):

packages/base/src/3dview/mainview.tsx
martinRenou: Looks like we need some kind of `Promise.all`

gjmooney: I added a `Promise.all` after the loop.

martinRenou (approved):
🚀

trungleduc (approved):

PR #353, opened 2024-04-24
Update 3D rendering logic

### LINKED ISSUE #335: STL normals are not correct, producing visual artifacts
```
diff --git a/packages/base/src/mainview.tsx b/packages/base/src/mainview.tsx
index d8c8bd0..c1717b3 100644
--- a/packages/base/src/mainview.tsx
+++ b/packages/base/src/mainview.tsx
@@ -745,7 +745,7 @@ export class MainView extends React.Component<IProps, IStates> {
       // it's too bad Three.js does not easily allow setting uniforms independently per-mesh
       const material = new THREE.MeshPhongMaterial({
         color,
-        side: THREE.DoubleSide,
+        // side: THREE.DoubleSide,
         wireframe: false,
         flatShading: false,
         clippingPlanes: this._clippingPlanes,
```

Using this diff to only render the "front faces" we see rendering issues:
Screenshot from 2024-03-13 12-38-30

This has to do with wrong normal computation or wrong node indexing.

trungleduc:
# Reference

Closes #335 

STL file renders correctly with single-side material

stl-fix

# Code changes

- Fix vertex order issue in `makeShapeFromMesh`.
- Move the vertex normals computation to ThreeJS.
- Improve the loading animation in the 3D view.
- Reduce edge size by half.
- Fix the issue of rendering many vertex due to the spread operator.


trungleduc:
please update snapshots!

packages/base/src/3dview/helpers.ts
```diff
@@ -167,20 +168,20 @@ export function buildShape(options: {
...
   });
 
   const geometry = new THREE.BufferGeometry();
+
   geometry.setIndex(triangles);
   geometry.setAttribute(
     'position',
     new THREE.Float32BufferAttribute(vertices, 3)
   );
-  geometry.setAttribute('normal', new THREE.Float32BufferAttribute(normals, 3));
```
martinRenou, line 183: ❤️

packages/base/src/3dview/helpers.ts
```diff
@@ -167,20 +168,20 @@ export function buildShape(options: {
...
+  geometry.computeVertexNormals();
```
martinRenou, line 184: Is this really needed? If nodes are ordered correctly, we can save a tiny bit of computation and let the GPU figure out the normals when rendering

martinRenou: Let's not block the PR for this, it's a detail

martinRenou (approved):
Thanks!

PR #355, opened 2024-05-07
Update jcad exporting mechanism 

trungleduc:
## Code changes

- Allows reusing `Export to JCAD` dialog for different file types
- Adds `STL` to `JCAD` export
- Fixes edge selection issue in the properties panel.


trungleduc:
please update snapshots!

PR #364, opened 2024-06-14
Include Notebook API into lite kernel deployment

martinRenou:


trungleduc:
why do we need to rebuild the extension and install from source? IIRC i downloaded the artifact from previous step and install the `whl`s directly

martinRenou:
Ah indeed I could install the whl directly, but that's not what's being done currently

trungleduc:
> Ah indeed I could install the whl directly, but that's not what's being done currently

it is done here https://github.com/jupytercad/jupytercad/blob/aaf0e39e9ce0e67877ee0c073d32dca808d6811d/.github/workflows/build.yml#L224 ?

martinRenou:
The only impact of this will be the installation of the labextensions in the build environment, then JupyterLite will pick up automatically those labextensions and include them in the lite build.

But it will not install the Python package in the kernel runtime.

trungleduc (approved):

martinRenou:
This is just me trying things, I believe the notebook API should be able to open document shared models as we're going through comms with ypywidgets and not using server component. But let's try!

martinRenou:
Closing as it's apparently not possible to make this work unless the tokens for collaboration are on.

martinRenou:
For the record, this is what I get:

Screenshot from 2024-06-27 12-54-52

There is no error in the console whatsoever.

trungleduc:
there is no 3d view in notebook?

martinRenou:
Yep, I'm not sure where it's blocking here:

Screenshot from 2024-06-27 13-17-29

trungleduc:
you might need to set up a dev version of `yjs-widgets` to debug it.

Issue #363, opened 2024-06-14
`ArchDetail.FCstd` example fails to open with `jupytercad==2.0.0a9`


martinRenou:
`ArchDetail.FCstd` example fails to open with `jupytercad==2.0.0a9` with the error message:

```
Error: Unknown error while creating Part::Extrusion: {"objects":[{"shape":"Part::FeaturePython","visible":false,"parameters":{"MoveBase":false,"BaseOffsetY":0,"Nodes":null,"IfcData":{"attributes":"{\"PredefinedType\": {\"name\": \"PredefinedType\", \"type\": \"IfcBuildingElementProxyTypeEnum\", \"is_enum\": true, \"enum_values\": [\"COMPLEX\", \"ELEMENT\", \"PARTIAL\", \"PROVISIONFORVOID\", \"PROVISIONFORSPACE\", \"USERDEFINED\", \"NOTDEFINED\"], \"value\": \"COMPLEX\"}, \"Tag\": {\"name\": \"Tag\", \"type\": \"IfcIdentifier\", \"is_enum\": false, \"enum_values\": []}}"},"ComputedLength":0,"Description":null,"Normal":null,"Additions":[],"Width":14,"Axes":[],"Material":null,"StandardCode":null,"FaceMaker":null,"Shape":"\nCASCADE Topology V1, (c) Matra-Datavision\nLocations 10\n1\n              1               0               0               0 \n              0               1               0               0 \n              0               0               1               0 \n1\n         …
```

I believe this has to do with the new error handling logic in jupytercad. Before, we ignored errors and produced empty object as output of operators if they failed (hence the `ArchDetail.FCstd` file was opening, only some shapes were not visible). Now, we raise an error and the full file fails to open.

We may need to rework the error handling and keep the old behavior. We should only handle errors upon dry runs. Also we should only do a dry-run upon user changes (to prevent users from breaking the file), not during the first file loads (as we don't seem to fully support all files).

PR #366, opened 2024-06-25
Fix dry-run logic so that it does not run on the first load of the file

martinRenou:
Fix https://github.com/jupytercad/JupyterCAD/issues/363

Some files are simply not fully supported by jupytercad yet, but we want to at least be able to partially open them (like the `ArchDetail.FCSTD` example).

This means we should not raise any error while loading those files in the worker and keep going if the resulting OCC shape of operators is not defined.

Error raising should only happen upon user interaction (creation of operators, changing of properties) so that we prevent users to create malformed files.

https://github.com/jupytercad/JupyterCAD/assets/21197331/896ac810-5533-4102-80f3-e27ced0b34c8

### DIFF
--- a/packages/occ-worker/src/actions.ts
+++ b/packages/occ-worker/src/actions.ts
@@ -63,11 +63,11 @@ function buildModel(
   return outputModel;
 }
 
-function loadFile(payload: { content: IJCadContent }): IDict | null {
+function loadFile(payload: { content: IJCadContent }, load_on_failure=false): IDict | null {
   const { content } = payload;
   const outputModel = buildModel(content);
 
-  const parser = new OccParser(outputModel);
+  const parser = new OccParser(outputModel, load_on_failure);
   const result = parser.execute();
   const postResult: IDict<IPostOperatorInput> = {};
   outputModel.forEach(item => {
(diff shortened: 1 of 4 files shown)


packages/occ-worker/src/actions.ts
```diff
@@ -63,11 +63,14 @@ function buildModel(
   return outputModel;
 }
 
-function loadFile(payload: { content: IJCadContent }): IDict | null {
+function loadFile(
+  payload: { content: IJCadContent },
+  load_on_failure = false
```
trungleduc, line 68:
```suggestion
  loadOnFailure = false
```

packages/occ-worker/src/occparser.ts
```diff
@@ -11,8 +11,11 @@ interface IShapeList {
 export class OccParser {
   private _shapeList: IShapeList[];
   private _occ: OCC.OpenCascadeInstance = (self as any).occ;
-  constructor(shapeList: IShapeList[]) {
+  private raise_on_failure: boolean;
```
trungleduc, line 14:
```suggestion
  private raiseOnFailure: boolean;
```

packages/occ-worker/src/actions.ts
```diff
@@ -63,11 +63,14 @@ function buildModel(
...
+): IDict | null {
   const { content } = payload;
   const outputModel = buildModel(content);
 
-  const parser = new OccParser(outputModel);
+  const parser = new OccParser(outputModel, load_on_failure);
```
trungleduc, line 73: I feel more natural to use the `loadOnFailure` parameter as the `parser. execute` parameter

trungleduc (changes requested):

packages/occ-worker/src/actions.ts
martinRenou: I've done too much Python lately :D Thanks

trungleduc (approved):

trungleduc:
Thanks!

PR #369, opened 2024-06-26
Improve deletion behavior

### LINKED ISSUE #339: Prevent delete objects being used in other operators
JupyterCAD allows deleting objects being used by other operators, this breaks the rendering chain. We should check for it before deleting the objects.

martinRenou:
Fix #339 

Removing objects now result in opening a dialog showing all the objects that will get deleted in the process if there are dependants operators.

https://github.com/jupytercad/JupyterCAD/assets/21197331/d68b6c3b-9ccd-41dd-9d84-c4b3e8dcdf5b

### DIFF
--- a/packages/schema/src/doc.ts
+++ b/packages/schema/src/doc.ts
@@ -147,7 +147,34 @@ export class JupyterCadDoc
     if (!obj) {
       return;
     }
-    this.transact(() => obj.set(key, value));
+
+    this.transact(() => {
+      // Special case for changing parameters, we may need to update dependencies
+      console.log('update ', key);
+      if (key === 'parameters') {
+        switch(obj.get('shape')) {
+          case 'Part::Cut': {
+            obj.set('dependencies', [value['Base'], value['Tool']]);
+            break;
+          }
+          case 'Part::Extrusion':
+          case 'Part::Fillet':
+          case 'Part::Chamfer': {
+            obj.set('dependencies', [value['Base']]);
+            break;
+          }
+          case 'Part::MultiCommon':
+          case 'Part::MultiFuse': {
+            obj.set('dependencies', value['Shapes']);
+            break;
+          }
+          default:
+            break;
+        }
+      }
+
+      obj.set(key, value);
+    });
   }
 
   getOption(key: keyof IJCadOptions): IDict | undefined {
--- a/packages/schema/src/doc.ts
+++ b/packages/schema/src/doc.ts
@@ -81,7 +81,32 @@ export class JupyterCadDoc
     return undefined;
   }
 
+  getDependants(name: string): string[] {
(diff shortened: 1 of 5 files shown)


SylvainCorlay:
This is great! I think this will make 2.0 a lot more solid in terms of user experience.

martinRenou:
There is still one thing to figure out, it's updating the dependencies entry in the object properties panel if Base/Tool/Shapes gets updated. Not sure there is a generic approach we can take here.

@trungleduc do you have ideas for that?

trungleduc:
we expand recursively the operator in the worker (https://github.com/jupytercad/JupyterCAD/blob/71e093ca9d979925396ee097d951a6d2ae438a44/packages/occ-worker/src/occapi/operatorCache.ts#L35). You can use this data to set the dependencies?

martinRenou:
https://github.com/jupytercad/JupyterCAD/assets/21197331/2db77473-9552-4941-94c9-f19983c99705

packages/schema/src/doc.ts
```diff
@@ -81,7 +81,32 @@ export class JupyterCadDoc
...
+      if ((obj.get('dependencies') || []).includes(name)) {
+        dependants.push(obj.get('name'));
+      }
+    }
+    let subDependants: string[] = [];
+    for (const dependant of dependants) {
+      subDependants = subDependants.concat(this.getDependants(dependant));
+    }
+    return dependants.concat(subDependants);
+  }
```
trungleduc, line 98: Depth-first search  can help here, but i'm not sure the following code compile

```js
  getDependants(name: string): string[] {
    const dependants: string[] = [];
    const dependantMap = new Map<string, Set<string>>();

    for (const obj of this._objects) {
      const deps: string[] = obj.get("dependencies") || [];
      const objName = obj.get("name");
      deps.forEach((dep) => {
        const currentSet = dependantMap.get(dep);
        if (currentSet) {
          currentSet.add(objName);
        } else {
          dependantMap.set(dep, new Set([objName]));
        }
      });
    }
    const selectedDeps = dependantMap.get(name);
    if (!selectedDeps) {
      return [];
    }
    while (selectedDeps.size) {
      const depsList = [...selectedDeps];
      depsList.forEach((it) => {
        dependants.push(it);
        selectedDeps.delete(it);
        dependantMap.get(id)?.forEach((newIt) => selectedDeps.add(newIt));
      });
    }

    return dependants;
  }
```

martinRenou: Thanks!

trungleduc: does it work? I did not check the result locally

trungleduc (approved):

trungleduc:
Thanks!

PR #371, opened 2024-06-26
Generate dropdowns for operands on operator properties

### LINKED ISSUE #370: Object properties panel not showing dropdown menus for operands
When creating an operator you can select the operands from dropdown menus, once created you can change the operands in the right panel but the UI is not a dropdown anymore, it's a text input.

martinRenou:
Screenshot from 2024-06-26 12-02-36

Fix #370

### DIFF
--- a/packages/base/src/panelview/objectproperties.tsx
+++ b/packages/base/src/panelview/objectproperties.tsx
@@ -293,10 +293,10 @@ class ObjectPropertiesReact extends React.Component<IProps, IStates> {
     if (form_schema) {
       const allObjects = this.props.cpModel.jcadModel
         ?.getAllObject()
-        .map(o => o.name);
-      if (allObjects && this.state.selectedObject) {
-        delete allObjects[allObjects.indexOf(this.state.selectedObject)];
-      }
+        .reduce((all, o) => {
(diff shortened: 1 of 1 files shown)


packages/base/src/panelview/objectproperties.tsx
```diff
@@ -288,11 +288,37 @@ class ObjectPropertiesReact extends React.Component<IProps, IStates> {
...
   render(): React.ReactNode {
+    // Fill form schema with available objects
+    const form_schema = this.state.schema;
+    if (form_schema) {
+      const allObjects = this.props.cpModel.jcadModel
+        ?.getAllObject()
+        .map(o => o.name);
+      if (allObjects && this.state.selectedObject) {
+        delete allObjects[allObjects.indexOf(this.state.selectedObject)];
+      }
```
trungleduc, line 299:
```suggestion
      const allObjects = this.props.cpModel.jcadModel
        ?.getAllObject()
        .reduce((all, o) => {
          o.name !== this.state.selectedObject && all.push(o.name);
          return all;
        }, []);
```

trungleduc: it's not shorter but i think it looks nicer

trungleduc (approved):

Issue #356, opened 2024-06-03
Extrusion produces wrong indexing -> visual artifacts


martinRenou:
Similar to https://github.com/jupytercad/jupytercad/issues/335 but for extrusion result

![Screenshot from 2024-06-03 10-50-00](https://github.com/jupytercad/jupytercad/assets/21197331/88678041-7c4e-4a90-aa76-c909d64db4a6)

trungleduc:
hmm, could you test it with a previous version, maybe before https://github.com/jupytercad/jupytercad/pull/353

martinRenou:
> hmm, could you test it with a previous version, maybe before https://github.com/jupytercad/JupyterCAD/pull/353 ?

I also reproduce the same-ish issue before that PR

martinRenou:
It looks like it was working in `2.0.0a0` and stopped working in `2.0.0a1`

trungleduc:
wow, there are only 3 commits between a0 and a1, and two are meaningless. 

![jcad](https://github.com/jupytercad/JupyterCAD/assets/4451292/6b57809a-138d-448f-9f9c-5dbd338884dc)

trungleduc:
actually, we are wrongly pinning the `jupytercad-*` package in the meta package. Installing `jupytercad` always gets the latest  `jupytercad-*` ones. This issue happens between a2 and a4, which contains the 3D refactoring pass. I will take a look at it

martinRenou:
Good catch!

trungleduc:
I managed to track it down into 2 issues:
- the front face and back face create additional meshes in the scene. https://github.com/jupytercad/JupyterCAD/blob/71e093ca9d979925396ee097d951a6d2ae438a44/packages/base/src/3dview/helpers.ts#L208 

![jcad0](https://github.com/jupytercad/JupyterCAD/assets/4451292/19dc2e84-be2f-424c-b0fb-4b15e898829b)

It looks like these faces are related to the clipping panel, they are introduced in #333 

- if I remove it, I get a better one

![jcad4](https://github.com/jupytercad/JupyterCAD/assets/4451292/5694d12a-b55d-45d2-9bbd-71ad681cf286)



it still has the face normal points to the inside, but I'm not sure if it's an issue or it's the behavior of OpenCascade, since if I extrude in the opposite direction, I get the face normal points to the outside.

[...]

PR #372, opened 2024-06-26
Draft: Function for checking if an object is solid

martinRenou:
Attempt to find a generic solution for fixing the display of non-solid shapes (see https://github.com/jupytercad/JupyterCAD/issues/356)


martinRenou:
I'll keep it there for future me to pick up, but I will fix https://github.com/jupytercad/JupyterCAD/issues/356 using a workaround for now

PR #382, opened 2024-07-20
Implement `Undo` & `Redo` keyboard shortcuts

arjxn-py:
This is a rough try of me in which I tried to introduce simple `keybindings` for `undo` & `redo`

## Summary of What I Did:
- Created `keybindings.json` in which I have defined keybindings for undo and redo.
- Loaded Keybindings inside Extension:
  - Modified the `addCommands()` function to load the keybindings.
  - Used a helper function `loadKeybindings` to register the keybindings with the command registry.

### DIFF
--- a/packages/base/src/commands.ts
+++ b/packages/base/src/commands.ts
@@ -9,6 +9,7 @@ import {
   ISelection,
   Parts
 } from '@jupytercad/schema';
+import { CommandRegistry } from '@lumino/commands';
 import { JupyterFrontEnd } from '@jupyterlab/application';
 import { showErrorMessage, WidgetTracker } from '@jupyterlab/apputils';
 import { ITranslator } from '@jupyterlab/translation';
@@ -33,6 +34,7 @@ import {
   chamferIcon,
   filletIcon
 } from './tools';
+import keybindings from './keybindings.json';
 import { JupyterCadPanel, JupyterCadWidget } from './widget';
 import { DocumentRegistry } from '@jupyterlab/docregistry';
 import { PathExt } from '@jupyterlab/coreutils';
@@ -638,6 +640,16 @@ const EXPORT_FORM = {
   }
 };
 
+function loadKeybindings(commands: CommandRegistry, keybindings: any[]) {
+  keybindings.forEach(binding => {
+    commands.addKeyBinding({
+      command: binding.command,
+      keys: binding.keys,
+      selector: binding.selector
+    });
+  });
+}
+
 /**
  * Add the FreeCAD commands to the application's command registry.
  */
@@ -661,6 +673,7 @@ export function addCommands(
     },
     execute: args => {
       const current = tracker.currentWidget;
+      console.log('REDO PRESSED')
 
       if (current) {
         return current.context.model.sharedModel.redo();
@@ -678,6 +691,7 @@ export function addCommands(
     },
     execute: args => {
       const current = tracker.currentWidget;
+      console.log('UNDO PRESSED')
 
       if (current) {
         return current.context.model.sharedModel.undo();
@@ -953,6 +967,7 @@ export function addCommands(
       await dialog.launch();
     }
   });
+  loadKeybindings(commands, keybindings);
 }
 
 /**
--- a/packages/base/src/keybindings.json
+++ b/packages/base/src/keybindings.json
@@ -1,0 +1,12 @@
+[
+    {
+      "command": "jupytercad:undo",
+      "keys": ["Accel Z"],
+      "selector": "body"
+    },
+    {
+      "command": "jupytercad:redo",
+      "keys": ["Accel Y"],
+      "selector": "body"
+    }
+  ]
--- a/packages/base/tsconfig.json
+++ b/packages/base/tsconfig.json
@@ -4,5 +4,5 @@
     "outDir": "lib",
     "rootDir": "src"
   },
-  "include": ["src/**/*", "src/schema/*.json", "src/_interface/*.json"]
+  "include": ["src/**/*", "src/schema/*.json", "src/_interface/*.json", "src/*.json"]
 }


packages/base/tsconfig.json
```diff
@@ -4,5 +4,5 @@
     "outDir": "lib",
     "rootDir": "src"
   },
-  "include": ["src/**/*", "src/schema/*.json", "src/_interface/*.json"]
+  "include": ["src/**/*", "src/schema/*.json", "src/_interface/*.json", "src/*.json"]
```
arjxn-py, line 7: I've temporarily added this as I've saved `keybindings.json` in `src/`, will be happy to move it to more valid location.

packages/base/src/commands.ts
```diff
@@ -678,6 +691,7 @@ export function addCommands(
     },
     execute: args => {
       const current = tracker.currentWidget;
+      console.log('UNDO PRESSED')
```
arjxn-py, line 694: Just temporarily added these couple of log statements to get help while testing. Will remove them.

packages/base/src/commands.ts
```diff
@@ -661,6 +673,7 @@ export function addCommands(
     },
     execute: args => {
       const current = tracker.currentWidget;
+      console.log('REDO PRESSED')
```
arjxn-py, line 676: I noticed `redo` is not working properly for me locally, I add a model -> I do some changes -> I undo them -> I try to redo, but it is not functioning properly.

I verified the same on [https://jupytercad.github.io/JupyterCAD/lab/index.html](https://jupytercad.github.io/JupyterCAD/lab/index.html) too.

I'd be happy to open an Issue for the same upon confirmation.

arjxn-py:
I studied [QuantStack/jupytergis#58](https://github.com/QuantStack/jupytergis/pull/58), i updated the selector to be the jupytercad-panel for undo & redo. Need one opinion- having `keybindings.json` is better or defining those keybindings inside the script only?

packages/base/src/commands.ts
arjxn-py: Cheers, I already that branch locally too.
Thanks a lot @trungleduc 🎉 💯

martinRenou:
> Edit: I personally feel that having a keybindings.json file is good because it is more modular.

Yep I think that looks good 👍🏽 Thanks

It looks like your PR is now conflicting with the main branch, due to the merge of the other PR

arjxn-py:
Resolving this branch asap.

packages/base/src/commands.ts
```diff
@@ -697,6 +698,7 @@ export function addCommands(
     },
     execute: args => {
       const current = tracker.currentWidget;
+      console.log('UNDO PRESSED');
```
arjxn-py, line 701:
```suggestion
```

packages/base/src/commands.ts
```diff
@@ -680,6 +680,7 @@ export function addCommands(
     },
     execute: args => {
       const current = tracker.currentWidget;
+      console.log('REDO PRESSED');
```
arjxn-py, line 683:
```suggestion
```

packages/base/src/keybindings.json
```diff
@@ -1,4 +1,14 @@
 [
+  {
+    "command": "jupytercad:undo",
+    "keys": ["Accel Z"],
+    "selector": ".jp-jupytercad-panel"
+  },
+  {
+    "command": "jupytercad:redo",
+    "keys": ["Accel Y"],
+    "selector": ".jp-jupytercad-panel"
```
martinRenou, line 10: Should we do as in the other PR and add these keyboard shortcuts for the tree view?

At some point it will be repetitive to copy paste shortcuts for both the main view and the tree view, ideally we would define the shortcut only once for the entire app. But let's figure that out in a later-on PR. I don't want to block this PR for this.

martinRenou (changes requested):

packages/base/src/keybindings.json
arjxn-py, line 10:
```suggestion
    "selector": "body"
```

packages/base/src/keybindings.json
```diff
@@ -1,4 +1,14 @@
 [
+  {
+    "command": "jupytercad:undo",
+    "keys": ["Accel Z"],
+    "selector": ".jp-jupytercad-panel"
```
arjxn-py, line 5:
```suggestion
    "selector": "body"
```

arjxn-py (commented):
If I define body as the selector, these keybindings enables for whole app.
It is fine in that case if we don't want to have key bindings irrespective of which component is focused.

packages/base/src/keybindings.json
martinRenou: Does that actually work? That's great if that works.

I believe the command should behave properly and only execute the command on the currently active JupyterCad editor, if there is one.

packages/base/src/keybindings.json
arjxn-py: > At some point it will be repetitive to copy paste shortcuts for both the main view and the tree view, ideally we would define the shortcut only once for the entire app. But let's figure that out in a later-on PR. I don't want to block this PR for this.

One option is to have `body` as the selector and the other option is to wrap the components we want to have effects of key bindings in a container and then use that as a standard selector.

The latter is more fine grained but will take some time to implement and test effectively. Which I'll be more than happy to look into as well.

martinRenou: If the body selector just works, I'm fine with it:

> I believe the command should behave properly and only execute the command on the currently active JupyterCad editor, if there is one.

packages/base/src/keybindings.json
arjxn-py: > Does that actually work? That's great if that works.
> 
> I believe the command should behave properly and only execute the command on the currently active JupyterCad editor, if there is one.

Yes that works & as far as I had tested it, it behaves properly too.

arjxn-py: Even I had `body` as selector before 😆 in [`5518848`](https://github.com/jupytercad/JupyterCAD/pull/382/commits/55188485171439ea819ad37a36d87f494de306ca)
I saw https://github.com/QuantStack/jupytergis/pull/58 & thought that having fine grained control would be nicer, but definitely this is something we can standardize later too

martinRenou: Sounds good to me :) Let's use the body selector for all commands for now then

arjxn-py: image
Verified again when editor is not enabled.

arjxn-py: > Let's use the body selector for all commands for now then

Done, Thanks for the instructions :)

SylvainCorlay:
Arg. It seems the redo keyboard shortcut does not work.

arjxn-py:
Hi @SylvainCorlay, I just tested these keyboard shortcuts locally and on [https://jupytercad.github.io/JupyterCAD/](https://jupytercad.github.io/JupyterCAD/) as well and it seems to be working as expected for me on both the places.


https://github.com/user-attachments/assets/0b7fe9d3-c080-4e51-9ef5-7a6bf9c6fd92

However on a sidenote, I'd like to mention that we had an issue with the redo function in the past, which has been fixed recently by #385, thanks to @trungleduc. So in case you're trying this out locally, I'd once suggest to make sure that the code is properly synced to upstream. But in case the issue still persists, please let us know about the same. Thanks a lot for following up on this.

SylvainCorlay:
Thanks for looking into it. The issue may be specific to the JupyterLite deployment.

PR #385, opened 2024-07-24
Improve shape metadata update logic

### LINKED ISSUE #384: Backend undo manager
In jupyter-ydoc v3 there will be an [undo manager](https://github.com/jupyter-server/jupyter_ydoc/blob/54642267dc309be03c754a626aa75e0ac6f1c964/jupyter_ydoc/ybasedoc.py#L36) that will allow to undo/redo operations on a shared model in the backend in a document-agnostic way. Every document inheriting from `YBaseDoc` will be responsible for adding to the undo manager's scope the shared types that it wishes to be included. For instance, for a `YNotebook`, the [cells are included](https://github.com/jupyter-server/jupyter_ydoc/blob/54642267dc309be03c754a626aa75e0ac6f1c964/jupyter_ydoc/ynotebook.py#L59) in the undo manager's scope.
For a [YJCad](https://github.com/jupytercad/JupyterCAD/blob/d526259792a8cba1ba1e2d39194544fd8a1abfd3/python/jupytercad_core/jupytercad_core/jcad_ydoc.py#L9) document, it looks like `self._yobjects` should be included?

cc @meriem-benismail

trungleduc:
Fixes #384

https://github.com/user-attachments/assets/a510b967-50b5-49c2-86f0-02fe36a7b08b

## Code change

- Shape metadata is updated at the same time as the shape parameters
- Add tests for Undo and Redo function


packages/base/src/commands.ts
```diff
@@ -49,6 +51,23 @@ export function newName(type: string, model: IJupyterCadModel): string {
   return name;
 }
 
+export async function dryRunCheck(options: {
```
arjxn-py, line 57: How about we create an interface for options: 
```javascript
interface DryRunOptions {
  jcadContent: IJCadContent;
  mainView: MainViewModel;
  requestedOperator: string;
}
```
& use it as `dryRunCheck(options: DryRunOptions)`?

packages/base/src/commands.ts
```diff
@@ -205,18 +224,21 @@ export async function executeOperator(
...
-      `The ${name} tool was unable to create the desired shape due to invalid parameter values. The values you entered may not be compatible with the dimensions of your piece.`
-    );
 
+  const dryRunResult = await dryRunCheck({
+    jcadContent: updatedContent,
+    mainView: current.content.currentViewModel,
+    requestedOperator: name
+  });
+  if (!dryRunResult) {
     return;
```
arjxn-py, line 237: Would it be nice to also throw an error here in case someone want to debug this in future?

arjxn-py (commented):
I've tested this branch locally & it works seamlessly.
I would also want to suggest a couple of very minor improvements if that's ok.
Thanks a lot @trungleduc

packages/base/src/commands.ts
trungleduc: I would start creating an interface once it's used twice.

packages/base/src/commands.ts
trungleduc: The `dryRunCheck` function has already raised the error.

packages/base/src/commands.ts
arjxn-py: Right, that's fair.

packages/base/src/commands.ts
arjxn-py: Oh yes. Oops
That descriptive error will be thrown and will return null. That makes sense. Thanks 😄

martinRenou (approved):
Thanks a lot @trungleduc ! And thanks for the review @arjxn-py 🚀

PR #388, opened 2024-07-27
Delete Object Keyboard Shortcut

### LINKED ISSUE #46: Delete key binding
Add a `Delete` key binding for removing shapes

arjxn-py:
Should fix #46 
Adds a `Ctrl`+`X` keybinding to delete object.


packages/base/src/keybindings.json
```diff
@@ -0,0 +1,7 @@
+[
+  {
+    "command": "jupytercad:removeObject",
+    "keys": ["Accel X"],
```
martinRenou, line 4: I feel like "Accel X" should be the keybinding for "cut", though we don't have any copy/paste logic available so I'm fine keeping this keybinding for now if you like 👍🏽 

Though should we also add "Delete" there too in the list?

packages/base/src/keybindings.json
```diff
@@ -0,0 +1,7 @@
...
+    "selector": ".jp-jupytercad-panel"
```
martinRenou, line 5: This results in the key binding to not be triggered if it's the left panel (object list) that has focus.

I wonder if we should add the keybinding to the `#main` application instead. Though I tested locally and that does not seem to work, I suppose it's because the command depends on the `WidgetTracker<JupyterCadWidget>` tracker, and that one only tracks the 3D view if I'm correct.

Let's not consider this a blocker for merging the PR, but let's keep track of this in an issue after merging.

martinRenou (changes requested):
Neat! Thanks.

A small issue I spotted, when selecting an edge the title of the dialog is kind-of wrong, because we cannot just delete an edge. So we should probably find the parent object of the edge and delete that parent instead.


https://github.com/user-attachments/assets/e8c22a8c-4b27-4882-a38f-e222a485aa14

packages/base/src/keybindings.json
arjxn-py: I see, i can try adding the valid className for left panel too and see if the union of `jupytercad-panel` & `leftpanel` works.

packages/base/src/keybindings.json
arjxn-py: Thanks, that makes much sense as xref:[QuantStack/jupytergis#68](https://github.com/QuantStack/jupytergis/pull/68/files#diff-cb12ddc72814b382343561b2b38b359975e1ea74b168abd4c0db0d4e6d9d0a2bR94-R99) also uses the `[Delete]` key for the same.

Also it sounds nice to have a feature request open for cut/copy/paste logic. I can open one if it's fine.

martinRenou: > I can open one if it's fine

Sounds good!

packages/base/src/keybindings.json
martinRenou: Sounds good!

martinRenou (approved):
Thanks!

arjxn-py:
> A small issue I spotted, when selecting an edge the title of the dialog is kind-of wrong, because we cannot just delete an edge.

Thanks for merging this @martinRenou, And sorry for the delay I was still looking into the edge issue. I'll open a PR upstream as soon as i manage to fix it and if not I'll definitely open an issue to keep track of it.

trungleduc:
Great! Thanks for working on it

PR #389, opened 2024-07-27
Position & Axis inputs in a row

arjxn-py:
I saw [QuantStack/jupytergis#72](https://github.com/QuantStack/jupytergis/issues/72) & [QuantStack/jupytergis#74](https://github.com/QuantStack/jupytergis/pull/74) & thought it might be good here as well. Feel free to close this if it's not :)

Before: 
image
After:
image

credits: @martinRenou

### DIFF
--- a/python/jupytercad_lab/style/base.css
+++ b/python/jupytercad_lab/style/base.css
@@ -152,6 +152,21 @@
   overflow: hidden;
 }
 
+.jpcad-property-panel .array-item-list {
+  display: flex;
+}
+
+.jpcad-property-panel .jp-SchemaForm .array-item:not(:last-child) {
+  margin-right: var(--jp-ui-font-size1);
+  border-bottom: none;
+  margin-bottom: unset;
+  padding-bottom: unset;
+}
+
+.jpcad-property-panel .jp-SchemaForm .array-item {
+  flex-grow: 1;
+}
+
 .jpcad-hidden-field {
   display: none;
 }


arjxn-py:
I also did wanted to add `x`, `y` & `z` input labels instead of description for Position & Axis but tbh I wasn't able to exactly determine where and how are we rendering this form. I did saw we're using `ObjectPropertiesForm` from `formbuilder` hence tried modifying: 
https://github.com/jupytercad/JupyterCAD/blob/4c85db3c1287445f3d24fbb06df5bf750cb299ab/packages/base/src/panelview/formbuilder.tsx#L96
And also tried updating schema accordingly in [`placement.json`](https://github.com/jupytercad/JupyterCAD/blob/main/packages/schema/src/schema/placement.json).
But no luck.

martinRenou (approved):
Great thanks! ❤️ 

> tbh I wasn't able to exactly determine where and how are we rendering this form

The form is rendered using https://github.com/rjsf-team/react-jsonschema-form and the https://github.com/deathbeds/jupyterlab-starters/tree/main/packages/jupyterlab-rjsf form renderer.
I don't know this code enough to know if/where we can define such labels. You're also right we have custom logic for rendering some inputs it seems, maybe @trungleduc would know?

Happy to merge this PR as-is for now though. Thanks!

trungleduc:
the position is defined as an array for float, we would need to convert it into an object of 3 keys `x`, `y`, `z` to render the label but I don't think it is worth the effort.

Issue #123, opened 2023-04-24
3D view options


martinRenou:
We should provide to the user some view options:
- [x] orthographic and perspective projections
- [x] wireframe view
- [x] clipping

martinRenou:
We could get inspired by what https://old.opencascade.com/content/web-3d-demo is doing in terms of UX

martinRenou:
Closed as resolved!

PR #390, opened 2024-07-28
Enable Toggling Wireframe View

arjxn-py:
Related to #123 
image
image


packages/base/src/3dview/mainview.tsx
```diff
@@ -1257,6 +1283,9 @@ export class MainView extends React.Component<IProps, IStates> {
             : 'unset'
         }}
       >
+        <button onClick={this.toggleWireframe}>
+            Toggle Wireframe
+          </button>
```
arjxn-py: Just temporarily added this button for testing. Further plan is to add this in toolbar and also have a key binding for the same.

packages/base/src/3dview/mainview.tsx
```diff
@@ -104,8 +105,29 @@ export class MainView extends React.Component<IProps, IStates> {
...
+      () => {
+        this._scene.traverse((child) => {
+          if (child instanceof THREE.Mesh) {
+            child.material.wireframe = this.state.wireframe;
+            child.material.needsUpdate = true;
+          }
+        });
+
+        this._renderer.render(this._scene, this._camera);
+        console.log(this.state.wireframe);
```
arjxn-py, line 128: For testing. Will remove this.

arjxn-py:
I am having some hard time to access the method across files lately as I'm trying to register a command to toggle wireframe now. But I'm gonna be on it more over the weekend and report back in any case.

arjxn-py:
In case you'd like to give any feedback on current changes in the meantime that could refactor the approach or something, please do. Thanks.

arjxn-py:
https://github.com/user-attachments/assets/9e8641d8-decc-4564-88d4-5dffa0bb9539

Now have a working button in toolbar. Woohoo.

arjxn-py:
Button icon is a draft, you can let me know if I should look for a different icon. Plus i can notice that all other icon's color is teal but the new added one's is gray, even though all the source icons have gray color in the start. I must be missing something here.
image

And if you'd like I'd be happy to add a key binding for toggling wireframe right away, just suggest me what combination of keys would you like to toggle it. Thanks.

martinRenou:
That's neat! I like the icon!

> Plus i can notice that all other icon's color is teal but the new added one's is gray

The is this `jp-icon3` CSS class in the icon: https://github.com/jupytercad/JupyterCAD/blob/main/packages/base/style/icon/box.svg?short_path=c332c67#L3 Maybe that's what you're missing?

> just suggest me what combination of keys would you like to toggle it

Should we follow what FreeCAD does here, if it does it?

Just one note, I doubt the cliplane effect will work well with the wireframe view. We can disallow enabling both at the same time for simplicity, I can also try to find time to help on this front.

arjxn-py:
> Should we follow what FreeCAD does here, if it does it?

OMW to look into if FreeCAD does it or not.

> Maybe that's what you're missing?

Thanks :dart:

> I doubt the cliplane effect

Sorry, which one is cliplane effect? is it clipping one?

martinRenou:
> Sorry, which one is cliplane effect? is it clipping one?

No worries at all! Yes it's the clipping one

arjxn-py:
> Should we follow what FreeCAD does here, if it does it?

FreeCAD has a shortcut `Ctrl+Space` to toggle wireframe view. I have also added the same in [`92340a6`](https://github.com/jupytercad/JupyterCAD/pull/390/commits/92340a6f2a0ff59da404721ea51bc7723bcbbf94)

> this jp-icon3 CSS class in the icon

This is also done.

> I doubt the cliplane effect will work well with the wireframe view. We can disallow enabling both at the same time for simplicity

I tried doing this in [`1b0b238`](https://github.com/jupytercad/JupyterCAD/pull/390/commits/1b0b23811805426c0a9e0244d317b6ccfa91f2f7) but both of those are still enabled irrespective of the active command when I tested. I'd welcome your suggestions as it can save some work.

martinRenou:
Thanks! I will try to find time to have a look at the clipplane issue

arjxn-py:
> Thanks! I will try to find time to have a look at the clipplane issue

Thanks a lot, i'll be more than happy to assist you wherever required.

arjxn-py:
I'm OMW to resolve these conflicts. As since `packages/base/src/widget.tsx` &#8594; `packages/base/src/widget.ts` in #392 , I'm taking a little time and will also test it locally before pushing :)

arjxn-py:
I was a little lost on it, Sorry.
However the build is passing for me locally now and It should pass on CI too hopefully, but I'm getting a couple of new error logs in the browser which seem unrelated to the changes of this PR.
```
main.js:155 Error: Shared module @jupyterlab/codeeditor doesn't exist in shared scope default
```
```
main.js:155 Error: Shared module @jupyterlab/completer doesn't exist in shared scope default
```
Although I'm looking into if I can fix it or catch something substantial. I'd also welcome to hear from you why this might be happening.

arjxn-py:
> do you see these errors when running the JupyterCAD custom UI through the jupyter cad command?

Yes, these logs are from browser console when I startup the extension after running `jupyter cad` command.

> I guess this makes sense, and it is since the introduction of the console in JupyterCAD https://github.com/jupytercad/JupyterCAD/pull/392, it may prevent the console to work, but it's not due to your PR in particular so I would not worry too much on this PR.

To also mention, this error also prevents me from opening the jcad editor and open any file. Hence it is causing breakage at the moment which I've also been looking into. Apologies that I forgot to mention it earlier.

martinRenou:
I'd invite you to use `jupyter lab` instead for now, it should work there.

But we should definitely look into a fix before making a release of this!

trungleduc:
> I'd invite you to use `jupyter lab` instead for now, it should work there.
> 
> But we should definitely look into a fix before making a release of this!

Indeed, I forgot to check the standalone app in the console PR

arjxn-py:
> I'd invite you to use `jupyter lab` instead for now, it should work there.
> 
> But we should definitely look into a fix before making a release of this!

Sure, I'll do it as suggested and get back with update in any case. Thanks.

packages/base/src/widget.ts
```diff
@@ -138,6 +140,12 @@ export class JupyterCadPanel extends SplitPanel {
     this._view.delete('axes');
   }
 
+  handleToggleWireframe = () => {
+    if (this.mainViewRef.current) {
```
martinRenou, line 144: When testing your branch I get a `TypeError: this.mainViewRef is undefined` at runtime here.

How is this `mainViewRef` supposed to be set?

arjxn-py: `mainViewRef` is supposed to be the reference to the `MainView` component & set like `private mainViewRef: React.RefObject<MainView>;`. However, I don't get this `TypeError: this.mainViewRef is undefined` but `TypeError: Cannot read properties of undefined (reading 'current')` when I try to toggle wireframe. Sorry about this, I'm on my way to look into this and confirm once it gets fixed.

packages/base/src/3dview/mainview.tsx
```diff
@@ -1340,3 +1362,14 @@ export class MainView extends React.Component<IProps, IStates> {
   private _contextMenu: ContextMenu;
   private _loadingTimeout: ReturnType<typeof setTimeout> | null;
 }
+
+export function toggleWireframe(this: MainView) {
```
trungleduc, line 1366: Why is function is defined outside of the class and not its method?

packages/base/src/3dview/mainview.tsx
```diff
@@ -1297,6 +1308,17 @@ export class MainView extends React.Component<IProps, IStates> {
     );
   }
 
+  public setWireframe(wireframe: boolean) {
```
trungleduc, line 1311:
``` suggestion
  setWireframe(wireframe: boolean) {
```

packages/base/src/widget.ts
trungleduc: the ref object should not be used outside of the object referenced. And you don't need to use the ref here. 

Just follow the approach we are using for toggling the axes or exploded view by updating the `this._view` object. This object is passed to `MainView` object and we listen for the changed signal at https://github.com/jupytercad/JupyterCAD/blob/ca98c7bde6aa5e29592bc4d6152e345400f5f68c/packages/base/src/3dview/mainview.tsx#L1095

arjxn-py: Thanks for the suggestion @trungleduc, I think my approach was quite novice while i was trying to make it work & I should've discussed more. 
I think this PR needs some rework which I'll be doing after this and would be following what you've suggested :)

packages/base/src/3dview/mainview.tsx
arjxn-py: I was finding it difficult to access this method outside this file, hence did this for the time being.
I'd be on my way to define it inside the class and then try to use it across. I believe I'd have questions which I would be asking definitely while I'm on it specifically. Thanks for reviewing this.

arjxn-py:
I'm happy to mark this PR as draft for the time i'm reiterating on it. Please feel free to correct me if I should not.

arjxn-py:
https://github.com/user-attachments/assets/dfede216-2d96-4a45-b839-936bfe8f8ca2

Toggling wireframe works again. Now with the approach suggested by @trungleduc, thanks 💐 
Just preventing toggling wireframe & clipPlane at the same time remains.

packages/base/src/3dview/mainview.tsx
arjxn-py: Marking this as resolved as now it is being handled inside `_onViewChanged`

packages/base/src/widget.ts
arjxn-py: > Just follow the approach we are using for toggling the axes or exploded view by updating the `this._view` object. This object is passed to `MainView` object and we listen for the changed signal at

Done :)
Thanks again.

martinRenou:
Thanks for the rebase and update! I will test it and see if I can make the clipping work with it

arjxn-py:
> Thanks for the rebase and update! I will test it and see if I can make the clipping work with it

Sounds good to me, Thanks. 
I've also tried to figure it out but trying. I'll update in case i'd be able to pull it off :)

arjxn-py:
I'm now able to prevent enabling clip plane while wireframe view is enabled. It was my bad that all this time I was trying to do in the wrong file :face_in_clouds:

arjxn-py:
https://github.com/user-attachments/assets/ad3ee08f-9207-4bf9-a398-55fc3b1fd4ce

I've been able to prevent enabling both wireframe and clip plane at the same time. However there's this one case which creates a bottleneck condition i.e. when clip plane is left enabled even though wireframe view had also been enabled. Although it's informative in the console (referring to the attached recording), Maybe I can disable setting the `clipView?.enabled === true` while wireframe is enabled.

martinRenou:
> Just one note, I doubt the cliplane effect will work well with the wireframe view. We can disallow enabling both at the same time for simplicity, I can also try to find time to help on this front.

Looks like I was wrong on this, the clip view effect seems to work well with your wireframe view without requiring more work. I wrongly suspected there would be some visual glitches.

I guess we can go ahead allowing both at the same time!

packages/base/src/3dview/mainview.tsx
```diff
@@ -1110,6 +1122,23 @@ export class MainView extends React.Component<IProps, IStates> {
         this._updateClipping();
       }
     }
+
+    if (change.key === 'wireframe') {
+      const wireframeEnabled = change.newValue as boolean | undefined;
+
+      if (wireframeEnabled !== undefined) {
+        this.setState({ wireframe: wireframeEnabled }, () => {
+          this._scene.traverse(child => {
```
martinRenou, line 1131: Can you try only setting wireframe to true/false on the materials of the group `this._meshGroup`? That way you would not apply the wireframe effect on the clipplane control/pointers etc. We suppose only want the actual objects we're visualizing to be wireframed.

packages/base/src/widget.ts
```diff
@@ -138,6 +137,20 @@ export class JupyterCadPanel extends SplitPanel {
...
+    return this._view.get('wireframe') as boolean;
+  }
+
+  set wireframe(value: boolean) {
+    if (this.clipView?.enabled === true) {
+      console.warn(
+        'Cannot toggle Wireframe view while Clip Plane is enabled. Please disable the Clip Plane first.'
+      );
+      return;
+    }
```
martinRenou, line 150: Sorry for suggesting this, it seems it actually works great without further action from us!

```suggestion
```

packages/base/src/3dview/mainview.tsx
```diff
@@ -1102,6 +1108,12 @@ export class MainView extends React.Component<IProps, IStates> {
     }
 
     if (change.key === 'clipView') {
+      if (this.state.wireframe === true) {
+        console.warn(
+          'Cannot enable Clip Plane while Wireframe view is active. Please disable Wireframe view first.'
+        );
+        return;
+      }
```
martinRenou, line 1116: Same here

```suggestion
```

martinRenou (commented):
Thanks a lot, this works great! I just have some minor comments otherwise we can go ahead with this

packages/base/src/3dview/mainview.tsx
```diff
@@ -829,7 +834,8 @@ export class MainView extends React.Component<IProps, IStates> {
             clientColor.g / 255,
             clientColor.b / 255
           )
-        : 'black'
+        : 'black',
+      wireframe: this.state.wireframe
```
martinRenou, line 838: I would not apply the wireframe on the 3D mouse pointers

```suggestion
        : 'black'
```

packages/base/src/3dview/mainview.tsx
```diff
@@ -305,7 +307,8 @@ export class MainView extends React.Component<IProps, IStates> {
           color: 'black',
           opacity: 0.2,
           transparent: true,
-          side: THREE.DoubleSide
+          side: THREE.DoubleSide,
+          wireframe: this.state.wireframe
```
martinRenou, line 311: Also not on the clipplane control object

```suggestion
          side: THREE.DoubleSide
```

arjxn-py:
Hi @martinRenou, thanks for the review.
I'll integrate it today as soon as possible. Apologies for the delay.

martinRenou:
> Apologies for the delay.

No need for apologies, and no worries for integrating the changes now, it can wait for next week! Huge thanks for this PR :)

packages/base/src/widget.ts
arjxn-py: No worries, it sounds good.

arjxn-py:
When i remove `// @ts-ignore` in [`18a2c7f`](https://github.com/jupytercad/JupyterCAD/pull/390/commits/18a2c7ffb077606ef85f8d0e2ee71c05074893a3), it gives error - Object is possibly 'null'. (Object referring to `this._meshGroup`)

Took me unexpectedly long time to figure out that `// @ts-ignore` can fix that :smiling_face_with_tear:
I was avoiding that from trying, feels a little like a fool now 🤣

packages/base/src/3dview/mainview.tsx
arjxn-py: Done in [`18a2c7f`](https://github.com/jupytercad/JupyterCAD/pull/390/commits/18a2c7ffb077606ef85f8d0e2ee71c05074893a3)

arjxn-py:
https://github.com/user-attachments/assets/65769754-b18b-4e90-a281-a6280557c771

Works as expected for me, pointer is not included in wireframe view and so is the clip plane.

packages/base/src/3dview/mainview.tsx
```diff
@@ -1120,7 +1120,8 @@ export class MainView extends React.Component<IProps, IStates> {
 
       if (wireframeEnabled !== undefined) {
         this.setState({ wireframe: wireframeEnabled }, () => {
-          this._scene.traverse(child => {
+          // @ts-ignore
+          this._meshGroup.traverse(child => {
```
martinRenou, line 1124: Let's try our best to not use `@ts-ignore` in the codebase. I guess you had an TypeScript error here because `this._meshGroup` could potentially be undefined?

In that case I would do:
```
if (this._meshGroup !== undefined) {
  // Do the traverse
  // ...
}
```

martinRenou (changes requested):

packages/base/src/3dview/mainview.tsx
arjxn-py: Thanks @martinRenou, done in [`ca807d3`](https://github.com/jupytercad/JupyterCAD/pull/390/commits/ca807d33c84b31d0dd6d765fc9354f3a35ba61f8)

martinRenou:
Bot please update snapshots

packages/base/src/3dview/mainview.tsx
```diff
@@ -1110,6 +1114,24 @@ export class MainView extends React.Component<IProps, IStates> {
         this._updateClipping();
       }
     }
+
+    if (change.key === 'wireframe') {
+      const wireframeEnabled = change.newValue as boolean | undefined;
+
+      if (wireframeEnabled !== undefined) {
+        this.setState({ wireframe: wireframeEnabled }, () => {
```
trungleduc, line 1122: Setting the state like that will erase all other states. Using a function with an old state would be better.

trungleduc (changes requested):

packages/base/src/3dview/mainview.tsx
arjxn-py: On it, Thanks for suggesting.

arjxn-py:
> Bot please update snapshots

Looks like bot is not behaving correctly, would something that's been done in https://github.com/geojupyter/jupytergis/pull/97 fix the issue?

arjxn-py:
Other checks are not triggered for some reason, maybe manually triggering will do?

martinRenou:
> Other checks are not triggered for some reason, maybe manually triggering will do?

This is because the latest commit is by the bot, and bot commits don't trigger Github actions jobs. 

I'll close and reopen to re-trigger everything.

martinRenou (approved):
Thanks!

trungleduc:
Thanks!

PR #392, opened 2024-07-30
Add python console to 3D view

trungleduc:
## User-facing changes

- New toolbar button to open the python console
- The JCAD document is loaded automatically into the `doc` variable

https://github.com/user-attachments/assets/fd1dfb61-a345-4ff8-8891-7344e9a79b5f

## Code changes

- The main view is now a split panel instead of a panel


martinRenou:
Cross referencing with https://github.com/QuantStack/jupytergis/issues/41 where we discussed with David about potentially saving the state of the console

trungleduc:
What is the benefit of keeping the console state in the document?

martinRenou:
I figured it was nice to see the history of executed code to see how the file was modified/what data was loaded etc. 

I've not played so much with the console in tools like blender/freecad, do they not store the history?

Not storing the history makes life simpler so if the benefit is small, let's not do it.

trungleduc:
> I figured it was nice to see the history of executed code to see how the file was modified/what data was loaded etc.

But if you re-execute these codes, you might mess up your CAD document since their effects are already applied to the doc.

> I've not played so much with the console in tools like blender/freecad, do they not store the history?

Nope, but their GUI is basically a shortcut to call Python APIs, so you can translate your GUI working session into a Python script.

packages/base/src/toolbar/usertoolbaritem.tsx
```diff
@@ -71,10 +71,21 @@ export class UsersItem extends React.Component<IProps, IState> {
...
+        newList.push(element);
+      }
+    }
+    return newList;
+  }
   render(): React.ReactNode {
     return (
       <div className="jpcad-toolbar-usertoolbar">
-        {this.state.usersList.map(item => {
+        {this.filterDuplicated(this.state.usersList).map(item => {
```
martinRenou, line 88: Nice! I've seen this issue once indeed when using the Python API next to the main view

martinRenou (approved):
Thanks! 

For a future PR, we may want to use the StateDB to store the state of the console in the main view (open/close, expanded/hidden)

PR #393, opened 2024-07-31
Delete parent object if edge is selected with `[Delete]` key binding

arjxn-py:
A follow up PR for #388

https://github.com/user-attachments/assets/aed584ee-e6f8-4067-8542-64ead7986bf2


martinRenou (approved):
Neat! Thanks

PR #395, opened 2024-08-11
Enable Damping for Smoother Model Movement

arjxn-py:
This PR introduces damping to the model controls in JupyterCAD to provide smoother and more intuitive movement when interacting with 3D models. The damping effect has been enabled using a damping factor of `0.15`

ScreenRecording2024-08-11175758-ezgif com-video-to-gif-converter


arjxn-py:
I've also doing work in the same direction in [`smooth-movement`](https://github.com/arjxn-py/JupyterCAD/tree/smooth-movement) branch of my fork where i also wanted to make the zoom smoother too.

I've tried with `trackBallControls` first which was not behaving 100% correctly, then also tried defining `customControls` in [this commit](https://github.com/arjxn-py/JupyterCAD/commit/5e5a7b2fe0fcd648ac986f74b4d1540bcffd48eb) but not much difference.

martinRenou (approved):
Thanks!

Issue #394, opened 2024-08-10
[ENH] Object Color Customization in JupyterCAD


arjxn-py:
Currently, JupyterCAD models have a single default color, which is gray. It would be highly beneficial to introduce a feature allowing users to change the color of individual objects within a model. This could be implemented using hex codes and a color picker, giving users the flexibility to customize the appearance of their models to suit their needs.

Proposed Solution:

- Introduce a color picker UI element in the right panel maybe.
- Allow users to input hex color codes in an input field in object properties form.
- Have an option to apply these colors to individual objects within the model.

arjxn-py:
I want to ask where are we setting the color of the objects. I ask because I've already looked into `mainview` but couldn't exactly find.

martinRenou:
Thanks for the suggestion! 

We have color support when loading FCSTD files (through the `guidata` entry in the model), but we don't allow editing those colors. Also we indeed don't provide this feature for the jcad format.

I believe we should be able to add a "Color" property in the objects schemas, similar to all objects having a "Placement": https://github.com/jupytercad/JupyterCAD/blob/main/packages/schema/src/schema/box.json#L23
By doing so, a color picker should be visible in the object properties form automatically (at least a text input box, but we could make it a color picker by tweaking rjsf similar to what we've done in JupyterGIS).

We would then need the `mainview` to support reading this color and applying it to the 3D mesh material.

I'm thinking we may ditch the guidata thingy for FCSTD files and set the color through that new "Color" property we would be introducing.

arjxn-py:
> I believe we should be able to add a "Color" property in the objects schemas, similar to all objects having a "Placement":

Thanks for the clarification, this sounds like a good start. I should give it a try & discuss more retrospectively.

martinRenou:
Closed as resolved, thanks @arjxn-py !

PR #397, opened 2024-09-02
Add Color Customization for JupyterCAD Models

arjxn-py:
This PR adds a new property "Color" in JupyterCAD objects, that allows users to change the color of individual objects within a model.
image

related to #394


trungleduc:
Thank you for working on it. Adding a new `Color` property would make it harder to maintain compatibility with the FreeCAD file format. Why don't we reuse the `guidata` logic?

martinRenou:
@trungleduc I suggested getting rid of guidata in here: https://github.com/jupytercad/JupyterCAD/issues/394

I believe we should do the guidata -> color/transparency translation in Python in our freecad handler. What do you think?

trungleduc:
> @trungleduc I suggested getting rid of guidata in here: #394
> 
> I believe we should do the guidata -> color/transparency translation in Python in our freecad handler. What do you think?

👍

martinRenou:
> I'm not sure about jcad -> fcstd

It should also be fairly easy, it is just a dictionary `{obj_name: options}` with options being `{ShapeColor: color, 'Visibility': visible}` https://github.com/jupytercad/JupyterCAD-FreeCAD/blob/main/jupytercad_freecad/freecad/loader.py#L48

arjxn-py:
https://github.com/user-attachments/assets/6843d942-c531-49ef-bffc-94c380b01cd6

Setting custom color [here](https://github.com/jupytercad/JupyterCAD/pull/397/files#diff-37a7f3f4d02e2bc4c4f40a612b75d24569d7c0c46420298e3281f4af6e29b7e6R1067) in `mainview` changes the color of the object but on first interaction it comes back to it's original color. 
I'm also a little unsure about how do i access the color property from object schema in `mainview`? I had been trying on that but I think it'd be very much helpful to have your suggestion for that. Thanks :)

martinRenou:
You would also need to set the color [here when building the shapes](https://github.com/jupytercad/JupyterCAD/blob/main/packages/base/src/3dview/helpers.ts#L151). This gets called whenever there is a change in the object (size, position etc).

You should be able to get the color property around here:
https://github.com/jupytercad/JupyterCAD/blob/main/packages/base/src/3dview/mainview.tsx#L593
with something like `obj.parameters.Color`.

Note that you only changed the schema for the box, you'd need to change all schemas that should define a color.

martinRenou:
> More context: I've been logging and observing guidata to see if it has the object metadata and hence the color property. But could only find :

We'll eventually get rid of guidata, maybe in this PR, so you should not depend on it.

arjxn-py:
Thanks a lot, this really helps 💐 

> Note that you only changed the schema for the box, you'd need to change all schemas that should define a color.

Yes, I'm planning to do that once I have a proof of concept for box ready if that sounds fine. I can do it now as well if you'd like :)

arjxn-py:
> > More context: I've been logging and observing guidata to see if it has the object metadata and hence the color property. But could only find :
> 
> We'll eventually get rid of guidata, maybe in this PR, so you should not depend on it.

Makes sense, I'll try to avoid using it, thanks for validating.

packages/base/src/3dview/helpers.ts
```diff
@@ -148,14 +157,14 @@ export function buildShape(options: {
     vInd += vertexCoorLength / 3;
   }
 
-  let color = DEFAULT_MESH_COLOR;
+  let color = objColor || DEFAULT_MESH_COLOR;
+  console.log(color);
```
arjxn-py, line 161: I am getting the valid hexcode in browser console (i.e. which I input in the objectForm)
But still trying as I've been unable to bring this custom color to the model.
I'd be glad to again have your suggestion if I'm missing something. 
It could be something simple too but it's just that I've been unable to catch it by now. 😕 


https://github.com/user-attachments/assets/04807c2e-1ee1-4754-88e4-6a3c12caa955

Once this proof of concept is done, hopefully it'd be a breeze for me to implement the same for other shapes too.

packages/base/src/3dview/mainview.tsx
```diff
@@ -617,7 +619,7 @@ export class MainView extends React.Component<IProps, IStates> {
           this._edgeMaterials.push(el.material);
           if (selectedNames.includes(el.name)) {
             this._selectedMeshes.push(el as any as BasicMesh);
-            el.material.color = SELECTED_MESH_COLOR;
+            el.material.color = objColor;
```
martinRenou, line 622: I would keep `SELECTED_MESH_COLOR` here, as it's the color of selected meshes (for highlighting).

Though I believe you may be missing another spot where to set the color. When we deselect objects we bring back the `DEFAULT_MESH_COLOR`. I would search through the codebase everywhere `DEFAULT_MESH_COLOR` is used and replace it with `objColor || DEFAULT_MESH_COLOR`

arjxn-py: Thanks a lot. I'm now able to give custom color to the box model. woohoo 

https://github.com/user-attachments/assets/710d8a37-85b6-490c-b7ff-1e53192f2250

However it is still returning to `DEFAULT_MESH_COLOR` when deselected at the moment, i'm looking into this assuming it is in [`_updateSelected`](https://github.com/jupytercad/JupyterCAD/blob/main/packages/base/src/3dview/mainview.tsx#L842) method, I'll make sure to ask again if I feel stuck. Thanks again :)

arjxn-py:
https://github.com/user-attachments/assets/fb52ba73-e185-4791-826f-45f0e3530b08

Works nicely with box for me 🎉 
now on the other shapes

arjxn-py:
> I have a small suggestion, you should be able to get a nice and shiny color picker instead of a text input if you change [the uiSchema](https://github.com/jupytercad/JupyterCAD/blob/main/packages/base/src/panelview/formbuilder.tsx#L128) for the form building logic.

Thanks @martinRenou, I'll do it and follow up :)

arjxn-py:
https://github.com/user-attachments/assets/26424d86-cd4a-434a-80aa-cbc943dff012

Works even better with color picker that I myself am impressed 🤣 
Thanks @martinRenou

arjxn-py:
Hi @martinRenou, I think it'd be great to have your input here about which other schemas I should add `Color` property to?

martinRenou (changes requested):
Looks great! Thank you 🚀 

I have some comments (some a bit nitpicking to be fair).

> which other schemas I should add Color property to?

You will want to add a Color property to `any`, and operators like `chamfer`, `cut`, `extrusion`, `fillet`, `fuse` etc.

--- 

Once you added the color property to operators, you will also want to keep the color information upon applying an operator that takes one input. e.g. with the chamfer here, the result should stay red:

https://github.com/user-attachments/assets/ea89c00d-4640-4aea-8ca3-8605ffb0b14e

---

Concerning operators that take multiple inputs, we'll also want to be somewhat smart about the color information we keep. e.g. for the `cut` operator we may want to keep the color information from the `Base` (the object that gets cut). e.g. from the video here, it would make sense that the result is red, not orange:

https://github.com/user-attachments/assets/4bf68640-e78f-4251-99fb-df545ad8e935

The `fuse` operator may be trickier, should we blend the colors to get a new one? Happy to get suggestions on this.

arjxn-py:
Thanks for the comprehensive review!
I'll be on it to add `Color` property to all the remaining required schemas.
And I plan to handle these operations one after another iteratively hence first one in my list would be `chamfer`, if that sounds alright :)

packages/base/src/3dview/mainview.tsx
```diff
@@ -720,7 +726,7 @@ export class MainView extends React.Component<IProps, IStates> {
     }
 
     const material = new THREE.MeshPhongMaterial({
-      color: DEFAULT_MESH_COLOR,
+      color: new THREE.Color(0xff0000),
```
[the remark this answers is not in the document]
arjxn-py: Oops! this one's from while debugging 😅, thanks for noticing

packages/base/src/3dview/mainview.tsx
```diff
@@ -839,20 +845,27 @@ export class MainView extends React.Component<IProps, IStates> {
...
+    const selectedMesh = this._meshGroup?.getObjectByName(
+      meshName
+    ) as BasicMesh;
+
+    if (selectedMesh && selectedMesh.material?.color) {
+      const newColor = new THREE.Color(newColorHex);
+      selectedMesh.material.color = newColor;
+      selectedMesh.userData.originalColor = newColor.clone();
+    }
+  }
```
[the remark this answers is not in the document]
arjxn-py: Sure, that makes sense. I'll be on it to include it inline

packages/base/src/3dview/helpers.ts
```diff
@@ -114,12 +114,21 @@ export function buildShape(options: {
   selected: boolean;
   isSolid: boolean;
   guidata?: IDict;
+  objColor?: THREE.Color | string | number;
```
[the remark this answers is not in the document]
arjxn-py: Yes, as far as I know `0xff0000` has type number.
But that being said, I guess it should also not be harm to remove it and make `objColor` a little restrictive, if needed.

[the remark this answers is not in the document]
martinRenou: Ah right! Thanks for the clarification. Fine leaving this as-is.

martinRenou:
> Skip those which do not have placement prop

Yes that sounds correct 👍🏽

arjxn-py:
> The `fuse` operator may be trickier, should we blend the colors to get a new one? Happy to get suggestions on this.

I do not have much idea about `fuse` operator at the moment to have an opinion. But I shall workaround it and then have a valid suggestion for the same. In the meantime, it should be great to have @trungleduc's suggestion on it too. Thanks :)

trungleduc:
In jupyterCAD, the boolean operator output is a new shape, I would use the new shape default color

martinRenou:
> In JupyterCAD, the boolean operator output is a new shape, I would use the new shape default color

It would look so much nicer to keep the properties of the original object though, from a user point of view

martinRenou:
I was actually thinking that doing that for the placement could be nice too

martinRenou:
That's what FreeCAD does:

https://github.com/user-attachments/assets/12e14208-35b5-4d69-bcf7-7e3a6a8c511e

arjxn-py:
> > In JupyterCAD, the boolean operator output is a new shape, I would use the new shape default color
> 
> It would look so much nicer to keep the properties of the original object though, from a user point of view

Yes, it could be tried to preserve the properties of the object after operation if not already done. As far as I anticipate, Once done for color, Same could be implemented for placement too.

martinRenou:
At least, I was thinking to automatically pre-fill the creation form with the `Base`'s color value for the operators that have a `Base`.

The user can always change the color in the form.

arjxn-py:
> That's what FreeCAD does:

looks like they're not blending but the cylinder's surface in contact to the cube has the color of cube. (could be trickier but first step should be to preserve the cylinder's color for us in the same operation in JCAD)

martinRenou:
> surface in contact to the cube has the color of cube. (could be trickier but first step should be to preserve the cylinder's color for us in the same operation in JCAD)

Ah yes indeed I missed that!

Unfortunately we don't have any logic to render mesh faces separately

arjxn-py:
> Unfortunately we don't have any logic to render mesh faces separately

Yes, this could be a long term goal as it might need a good amount of refactoring too. 

Though I saw these (might not be totally related to our case but somewhat similar) - 
- [Changing color to the face of the cube when hover over occurs](https://discourse.threejs.org/t/changing-color-to-the-face-of-the-cube-when-hover-over-occurs/18100)
- [How to highlight a cube face on hover? THREE.js](https://discourse.threejs.org/t/how-to-highlight-a-cube-face-on-hover-three-js/27775)

martinRenou:
Yes that would be way too much refactoring. I'm not even sure we'd reach that point ever in JupyterCAD.

trungleduc:
> I was actually thinking that doing that for the placement could be nice too


> Yes, it could be tried to preserve the properties of the object after operation if not already done. As far as I anticipate, Once done for color, Same could be implemented for placement too.

No, placement should default to the zeros unless users modify it. The output of the boolean operator has already taken into account the placements of input objects. 

The placement is used to translate and rotate the local axes of an object in the global axes.

martinRenou:
> placement should default to the zeros unless users modify it. The output of the boolean operator has already taken into account the placements of input objects.

Right 👍🏽

arjxn-py:
Cut: 

https://github.com/user-attachments/assets/de7bd57a-eba5-4d3d-88aa-13e5049ac8bc

Union:

https://github.com/user-attachments/assets/566e3815-72b7-4c87-9047-13b47c3a3f93

Intersection: 

https://github.com/user-attachments/assets/20b0652f-c986-495f-b8a9-2f4f3dca58e7

arjxn-py:
I've given baseModel's color after each operation, feel free to let me know if that's not user friendly as I anticipated that it is.
These colors can always be updated.

arjxn-py:
Bot please update snapshots!

arjxn-py:
Oh! Oops, looks like bot is working(I thought it was not as it took much time)
But I think #400 should still be relevant, if not feel free to close :)

packages/base/src/commands.ts
```diff
@@ -257,11 +257,14 @@ const OPERATORS = {
...
       const sel1 = getSelectedMeshName(selected, 1);
+      const baseName = sel0 || objects[0].name || '';
+      const baseModel = model.sharedModel.getObjectByName(baseName);
       return {
         Name: newName('Cut', model),
-        Base: sel0 || objects[0].name || '',
+        Base: baseName,
         Tool: sel1 || objects[1].name || '',
         Refine: false,
+        Color: baseModel?.parameters?.Color,
```
martinRenou, line 267: Maybe provide the default value if it's not defined
```suggestion
        Color: baseModel?.parameters?.Color || DEFAULT_MESH_COLOR,
```

martinRenou: Same comment applies to other

packages/base/src/3dview/mainview.tsx
```diff
@@ -622,6 +624,17 @@ export class MainView extends React.Component<IProps, IStates> {
...
+        if (objColor && mainMesh?.material?.color) {
+          const selectedMesh = this._meshGroup?.getObjectByName(
+            mainMesh.name
+          ) as BasicMesh;
+          if (selectedMesh && selectedMesh.material?.color) {
+            const newColor = new THREE.Color(objColor);
+            selectedMesh.material.color = newColor;
+            selectedMesh.userData.originalColor = newColor.clone();
+          }
+        }
```
martinRenou, line 637: This is overwriting the initialization color of selected meshes (done line 622 of your file). 

You probably need to put this code here in and `else` fallback for the `if` condition line 620.

This is what causes the visual regression.

martinRenou (changes requested):

packages/base/src/commands.ts
arjxn-py: On it.

arjxn-py:
Bot please update snapshots!

arjxn-py:
> Bot please update snapshots!

Looks like bot is angry again 😭

martinRenou:
Ah right, the bot can only run if the build workflow passed already, because it reuses the build artifacts instead of making a custom build. I'll restart it now, that should be enough.

martinRenou:
Ah I know what's wrong. We need this https://github.com/geojupyter/jupytergis/blob/main/.github/workflows/update_galata_references.yaml#L79

arjxn-py:
> Ah I know what's wrong. We need this https://github.com/geojupyter/jupytergis/blob/main/.github/workflows/update_galata_references.yaml#L79

I see, Should I send a PR?

martinRenou:
Please do! It cannot be done in this PR because the bot runs from `main`, so we need to update `main` to apply

martinRenou:
It seems the bot did pick up the change https://github.com/jupytercad/JupyterCAD/actions/runs/10740182227/job/29787656302 (I see `Allow forks: true`) but it's still not happy

martinRenou:
Yep https://github.com/jupytercad/JupyterCAD/actions/runs/10740182227/job/29787926016 CI needs to have finished before the bot can run

arjxn-py:
Bot please update snapshots! :face_with_thermometer:

arjxn-py:
> Yep https://github.com/jupytercad/JupyterCAD/actions/runs/10740182227/job/29787926016 CI needs to have finished before the bot can run

Ahh! I missed this comment. Sorry
Hope it doesn't mess it up more

martinRenou:
It's fine, the second run will just cancel eventually

arjxn-py:
Updating branch w.r.t. `main` should trigger CI fully, should i do it now?

packages/base/src/3dview/mainview.tsx
```diff
@@ -619,9 +621,27 @@ export class MainView extends React.Component<IProps, IStates> {
...
+        if (!selected && objColor && mainMesh?.material?.color) {
+          const selectedMesh = this._meshGroup?.getObjectByName(
+            mainMesh.name
+          ) as BasicMesh;
+          if (selectedMesh && selectedMesh.material?.color) {
+            const newColor = new THREE.Color(objColor);
+            selectedMesh.material.color = newColor;
+            selectedMesh.userData.originalColor = newColor.clone();
+          }
+        }
```
martinRenou, line 644: Isn't it a duplication with what's just above? Is this some leftover from the old implementation we should remove?

packages/base/src/3dview/mainview.tsx
```diff
@@ -619,9 +621,27 @@ export class MainView extends React.Component<IProps, IStates> {
             this._selectedMeshes.push(el as any as BasicMesh);
             el.material.color = SELECTED_MESH_COLOR;
             el.material.linewidth = SELECTED_LINEWIDTH;
+          } else {
+            // Apply objColor for non-selected meshes
+            if (objColor && el.material?.color) {
+              const newColor = new THREE.Color(objColor);
+              el.material.color = newColor;
+              el.userData.originalColor = newColor.clone();
```
martinRenou, line 629: I feel like we should save `userData.originalColor` in all cases (whether it's selected or not)?

martinRenou (commented):
Thanks! I have a couple more comments.

packages/base/src/3dview/mainview.tsx
arjxn-py: Yes, I shall remove it. Thanks for noticing this. it's from one of the tries I was doing to have color.

packages/base/src/3dview/mainview.tsx
arjxn-py: Sure, on it.

packages/base/src/3dview/mainview.tsx
arjxn-py: Done in [`d952a76`](https://github.com/jupytercad/JupyterCAD/pull/397/commits/d952a760f8b8b0aba2c221ceaf7f9af659eb23c2)

arjxn-py:
image
Edges are no longer black or `DEFAULT_EDGE_COLOR`.

packages/base/src/3dview/mainview.tsx
```diff
@@ -619,6 +621,14 @@ export class MainView extends React.Component<IProps, IStates> {
             this._selectedMeshes.push(el as any as BasicMesh);
             el.material.color = SELECTED_MESH_COLOR;
             el.material.linewidth = SELECTED_LINEWIDTH;
+            el.userData.originalColor = el.material.color.clone();
```
martinRenou, line 624: Shouldn't it be the following? I thought `el.userData.originalColor` was used to store the mesh color as defined in the model?

```suggestion
            el.userData.originalColor = new THREE.Color(objColor);
```

martinRenou:
> Honestly, it personally looks good to me. What do you think?

I'm not sure about this change. This makes edges less visible, but those are clickable so they should be easily spotted by the user.

packages/base/src/3dview/mainview.tsx
arjxn-py: yes, my bad. Sorry.

arjxn-py:
> I'm not sure about this change. This makes edges less visible, but those are clickable so they should be easily spotted by the user.

Which one do you prefer though?

martinRenou:
Visually I prefer them in your screenshot, from the UX point of view I prefer them black. 

Though once the object is black it won't be visible anymore.

We may want a smart way to compute the edge color depending on the mesh color so that it's highlighted enough for us to see them.

martinRenou:
Let's not do that in this PR though

martinRenou (approved):
Thanks a lot! 🚀 

As a follow-up:
- get rid of guidata usage in the codebase (we will eventually need to introduce an "Visible" property to the objects)
- (Optional) figure out a smart way to compute the edge colors so that they are still visible but more pretty (make use of d3-color to make the color darker/brighter https://d3js.org/d3-color#color_darker)

arjxn-py:
> * get rid of guidata usage in the codebase (we will eventually need to introduce an "Visible" property to the objects)
> * (Optional) figure out a smart way to compute the edge colors so that they are still visible but more pretty (make use of d3-color to make the color darker/brighter https://d3js.org/d3-color#color_darker)

Thanks a lot @martinRenou, I'll be on these :)

PR #398, opened 2024-09-02
Fix bot behaviour to update snapshots

arjxn-py:
Same as https://github.com/geojupyter/jupytergis/pull/97


martinRenou (commented):
Thanks!

PR #402, opened 2024-09-09
Fix python console in the standalone app

### LINKED ISSUE #399: BUG: Error when Running JupyterCAD through `jupyter cad` Command
When running JupyterCAD through the `jupyter cad` command in the terminal, the following errors are encountered in the browser logs, and the JupyterCAD editor fails to open:

```
Error: Shared module @jupyterlab/codeeditor doesn't exist in shared scope default
```
```
Error: Shared module @jupyterlab/completer doesn't exist in shared scope default
```
However, this issue does not occur when JupyterCAD is accessed through the `jupyter lab` command. In that case, the editor opens successfully without any errors.

trungleduc:
- Add missing tokens to make python console work in the standalone app
- Closes #399
- Needs https://github.com/QuantStack/yjs-widgets/pull/14


martinRenou (approved):
Thanks!

martinRenou (approved):
Thanks!

arjxn-py:
Thanks @trungleduc, I got a bit late on picking #399 up 😄

trungleduc:
> Thanks @trungleduc, I got a bit late on picking #399 up 😄

don't worry, it is more complicated than it sounds

PR #404, opened 2024-09-09
Save jupyterlite build artifact

trungleduc:


arjxn-py (approved):
Thanks @trungleduc

.github/workflows/build.yml
```diff
@@ -235,6 +235,13 @@ jobs:
...
+        if: github.ref != 'refs/heads/main'
+        uses: actions/upload-artifact@v4
+        with:
+          path: ./lite/dist
+
+      - name: Upload github-pages artifact
+        if: github.ref == 'refs/heads/main'
         uses: actions/upload-pages-artifact@v1
         with:
           path: ./lite/dist
```
martinRenou, line 247: Are those lite build artifacts not already available through the upload with `actions/upload-pages-artifact@v1`?

trungleduc: yeah, I didn't realize that I could set the retention days for this action, the default value is just 1 day. Updating now

martinRenou (approved):

PR #405, opened 2024-09-10
Replace toggle wireframe button to the right

arjxn-py:
image


martinRenou (approved):
Thanks!

As a follow-up, it would also be nice to have a "selected" state on the button that would allow the user to see if the wireframe was toggled or not from the toolbar.

arjxn-py:
Locally when I'm compiling JCAD now, i'm needing to have to run `jlpm install` first as if not it gives me missing package error - 
```
Internal Error: @jupyterlab/apputils@npm:~4.0.0: This package doesn't seem to be present in your lockfile; run "yarn install" to update the lockfile
    at q0.getCandidates (C:\Users\Arjun\miniconda3\envs\jlab\Lib\site-packages\jupyterlab\staging\yarn.js:435:5145)
    at xf.getCandidates (C:\Users\Arjun\miniconda3\envs\jlab\Lib\site-packages\jupyterlab\staging\yarn.js:391:1264)
    at C:\Users\Arjun\miniconda3\envs\jlab\Lib\site-packages\jupyterlab\staging\yarn.js:439:7693
    at ff (C:\Users\Arjun\miniconda3\envs\jlab\Lib\site-packages\jupyterlab\staging\yarn.js:390:11070)
    at ge (C:\Users\Arjun\miniconda3\envs\jlab\Lib\site-packages\jupyterlab\staging\yarn.js:439:7673)
    at async Promise.allSettled (index 2)
    at async to (C:\Users\Arjun\miniconda3\envs\jlab\Lib\site-packages\jupyterlab\staging\yarn.js:390:10398)
```

However it fixes itself after i run `jlpm install` but it also modifies `yarn.lock` file. So should I push those changes or avoid?

arjxn-py:
> it would also be nice to have a "selected" state on the button that would allow the user to see if the wireframe was toggled or not from the toolbar.

Sounds fair, I'll look into that.
But as we have one more option -  do you prefer having the property `wireframe` in the schema or having the state for the button as you suggested? I've no opinion on this as both seems fine to me.

martinRenou:
> However it fixes itself after i run jlpm install but it also modifies yarn.lock file. So should I push those changes or avoid?

Yep, it sounds like that should be committed 👍🏽 

>  do you prefer having the property wireframe in the schema or having the state for the button as you suggested?

Indeed it would be nice to save the wireframe state in the schema, and the wireframe button would display if it's toggled or not 👍🏽 Also let's get the clip-plane button behave the same as your wireframe one, I prefer it that way personally and that would be more homogeneous in behavior. I'll open an issue to track this.

martinRenou:
x-ref https://github.com/jupytercad/JupyterCAD/issues/406

PR #407, opened 2024-09-10
Update dependencies

arjxn-py:
Discussed in #405 
Changes are auto-generated with `jlpm install` locally

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -840,7 +840,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@jupyter/ydoc@npm:^0.3.4 || ^1.0.2, @jupyter/ydoc@npm:^1.0.0, @jupyter/ydoc@npm:^1.1.0-a0":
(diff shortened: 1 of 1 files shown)


yarn.lock
```diff
@@ -1224,32 +1224,32 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@jupyterlab/apputils@npm:^4.0.0":
-  version: 4.3.5
-  resolution: "@jupyterlab/apputils@npm:4.3.5"
+"@jupyterlab/apputils@npm:~4.0.0":
+  version: 4.0.0
```
martinRenou, line 1228: I'm surprised this get downgraded.

martinRenou (commented):
I'm surprised by one of the changes, also when I try `jlpm install` locally it does not seem to change the lock file for me. I'm not sure what's different in your setup

yarn.lock
arjxn-py: So am I, it doesn't look right to me too.

arjxn-py: I think I shall try troubleshooting this issue for sometime locally first and then get back

arjxn-py:
Oh, it works. 
I guess it was an issue locally while switching branches, seems like `package.json` was incorrectly overwritten. Apologies.

martinRenou:
No worries, thanks for checking!

PR #408, opened 2024-09-10
Little darker shade of color for edges

arjxn-py:
Follow up for #394 & #397
image
image
image

### DIFF
--- a/packages/base/src/3dview/mainview.tsx
+++ b/packages/base/src/3dview/mainview.tsx
@@ -615,7 +615,9 @@ export class MainView extends React.Component<IProps, IStates> {
 
         // Save original color for the main mesh
         if (mainMesh.material?.color) {
-          const originalMeshColor = new THREE.Color(objColor || DEFAULT_MESH_COLOR);
+          const originalMeshColor = new THREE.Color(
+            objColor || DEFAULT_MESH_COLOR
+          );
           if (!selected) {
             mainMesh.material.color = originalMeshColor;
           }
--- a/packages/base/src/3dview/mainview.tsx
+++ b/packages/base/src/3dview/mainview.tsx
@@ -617,17 +617,16 @@ export class MainView extends React.Component<IProps, IStates> {
         }
         edgesMeshes.forEach(el => {
           this._edgeMaterials.push(el.material);
+          const originalEdgeColor = new THREE.Color(objColor).multiplyScalar(0.7);
           if (selectedNames.includes(el.name)) {
             this._selectedMeshes.push(el as any as BasicMesh);
             el.material.color = SELECTED_MESH_COLOR;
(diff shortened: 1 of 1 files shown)


arjxn-py:
There's a bug I encountered while changing object color, I'm on it fix that.

https://github.com/user-attachments/assets/486e3cd9-57a5-404e-a1c0-2472e62d11a7

arjxn-py:
When i change the color of mesh while selecting the whole mesh, the `originalColor` goes undefined here - 
https://github.com/jupytercad/JupyterCAD/blob/97a4abdce3e67b57574def4f208678efca5837f3/packages/base/src/3dview/mainview.tsx#L852-L864

However, this is not the case when i do it while selecting the edge as you can also notice in the clip in the above comment.

arjxn-py:
Solved 🎉

packages/base/src/3dview/mainview.tsx
```diff
@@ -612,22 +612,35 @@ export class MainView extends React.Component<IProps, IStates> {
...
           this._edgeMaterials.push(el.material);
+          const originalEdgeColor = new THREE.Color(objColor).multiplyScalar(
+            0.7
+          );
           if (selectedNames.includes(el.name)) {
             this._selectedMeshes.push(el as any as BasicMesh);
             el.material.color = SELECTED_MESH_COLOR;
             el.material.linewidth = SELECTED_LINEWIDTH;
-            el.userData.originalColor = new THREE.Color(objColor);
+            el.userData.originalColor = originalEdgeColor.clone();
```
arjxn-py, line 639: I believe that `new THREE.Color(originalEdgeColor)` & `originalEdgeColor.clone()` serves the same puspose, hence I felt like using this instead of again defining the color.

packages/base/src/3dview/mainview.tsx
```diff
@@ -612,22 +612,35 @@ export class MainView extends React.Component<IProps, IStates> {
...
+        // Save original color for the main mesh
+        if (mainMesh.material?.color) {
+          const originalMeshColor = new THREE.Color(
+            objColor || DEFAULT_MESH_COLOR
+          );
+          if (!selected) {
+            mainMesh.material.color = originalMeshColor;
+          }
+          mainMesh.userData.originalColor = originalMeshColor.clone();
+        }
```
martinRenou, line 625: Aren't we doing this already in `buildShape` called above?

arjxn-py:
https://github.com/user-attachments/assets/3b07574a-ca92-4fe0-a205-c98347cc35ce

martinRenou (commented):
Looks great! 🚀 

Now the blue-ish color for the selection looks really off to me. We should think of a better way to highlight selected objects.

trungleduc (approved):

trungleduc:
Thanks!

PR #412, opened 2024-09-11
Smart color handling for edges based upon mesh luminance

### LINKED ISSUE #411: Lighter edges for dark objects
We don't see edges for pitch black objects, we could fix that by checking if the objects are dark enough and rendering their edges lighter.

As discussed in https://github.com/jupytercad/JupyterCAD/pull/408#discussion_r1751885740

arjxn-py:
image
image
image
image

Should close #411


martinRenou (approved):
Thanks! 🚀

PR #413, opened 2024-09-12
Remove the guidata option

### LINKED ISSUE #410: Remove the guidata option
We should get rid of the `guidata` entry from the schema and everywhere in the codebase.

Instead we should introduce a new `Visible` property on all shapes, and a `Color` property ([done](https://github.com/jupytercad/JupyterCAD/pull/397)).

This is a breaking change and we'll have to update jupytercad-freecad accordingly.

arjxn-py:
Should potentially close #410


arjxn-py:
Bot please update snapshots!

martinRenou (changes requested):
Thanks for starting this!

I've not looked too deeply into the implementation yet, but it looks like this PR is breaking the some logic upon the application of a boolean operator. 
When creating two initial shapes, then applying a cut operator on them, the two initial shapes are not set hidden anymore.
Also when trying to hide them manually from the left panel seems to not work reliably.

I wonder if this is the cause of the visual regression the CI has seen.

arjxn-py:
> I've not looked too deeply into the implementation yet, but it looks like this PR is breaking the some logic upon the application of a boolean operator.
> When creating two initial shapes, then applying a cut operator on them, the two initial shapes are not set hidden anymore.
> Also when trying to hide them manually from the left panel seems to not work reliably.

I noticed this after your comment, i'm looking into why this is happening. Thanks for pointing it out.

packages/base/src/commands.ts
```diff
@@ -83,7 +83,9 @@ export function setVisible(
   if (object) {
     object.visible = value;
```
trungleduc, line 84: Why the `visible` value here does not work?

trungleduc:
color-error
This error here is related?

arjxn-py:
> This error here is related?

Might be related to #397, have to look into it. Thanks a lot for letting me know

arjxn-py:
Bot please update snapshots!

packages/base/src/commands.ts
arjxn-py: It works and as far as I can notice, it updates the visible property in the `.jcad` document, however it doesn't instantly updates the state of the model hence I have to also use sharedModel to directly update the state on the fly.

arjxn-py:
> > This error here is related?
> 
> Might be related to #397, have to look into it. Thanks a lot for letting me know

Hi @trungleduc, I tried reproducing this on my end locally but it seems like it works fine for me. However I'd want to confirm a couple of things - 
- Did you select any object before clicking intersection? (As it takes the color of that object)
- Do you also get `commands.js:262 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'value')` error when you try to do intersection with no obect or a single object? (This should be a valid error but maybe we could handle it better.)

arjxn-py:
https://github.com/user-attachments/assets/5149bc12-ca48-44b9-aced-3a7978d6f9c3

- Visibility toggling works fine
- Only result is visible after operation

trungleduc:
yeah, I can not reproduce it anymore, thanks for double-checking it!

packages/base/src/commands.ts
trungleduc: something like this is enough, `updateObjectByName` does check for object existence already

```ts
export function setVisible(
  sharedModel: IJupyterCadDoc,
  name: string,
  value: boolean
) {
  sharedModel.updateObjectByName(name, {
    data: { key: 'visible', value }
  });
}
```

packages/base/src/panelview/objecttree.tsx
```diff
@@ -421,24 +407,18 @@ class ObjectTreeReact extends React.Component<IProps, IStates> {
...
+                          const obj = this.getObjectFromName(objectId);
+                          if (obj) {
+                            obj.visible = !obj.visible;
+
+                            const sharedModel =
+                              this.props.cpModel.jcadModel?.sharedModel;
+                            if (sharedModel) {
+                              sharedModel.updateObjectByName(objectId, {
+                                data: { key: 'visible', value: obj.visible }
+                              });
```
trungleduc, line 419:
```suggestion
                          if (obj) {
                            const sharedModel =
                              this.props.cpModel.jcadModel?.sharedModel;
                            if (sharedModel) {
                              sharedModel.updateObjectByName(objectId, {
                                data: { key: 'visible', value: !obj.visible }
                              });
```

python/jupytercad_lab/jupytercad_lab/notebook/cad_document.py
```diff
@@ -166,31 +165,55 @@ def remove_annotation(self, annotation_id: str) -> None:
...
+        :param updated_obj: The updated object data.
+        """
+        objects = self._options.get("objects", [])
+
+        for i, obj in enumerate(objects):
+            if obj["name"] == object_name:
+                objects[i] = updated_obj
+                break
+
+        self._options.set(transaction, "objects", objects)
```
trungleduc, line 197: Did you try this method? I'm not sure it will work since `pycrdt.Map` does not have the `set` method. Let's remove the Python bits and do it in a follow-up PR.

trungleduc (changes requested):

packages/base/src/commands.ts
arjxn-py: Done in [`da3bac8`](https://github.com/jupytercad/JupyterCAD/pull/413/commits/da3bac8f6d773cb37cf0ec3ff9906fd172a71b85)

python/jupytercad_lab/jupytercad_lab/notebook/cad_document.py
arjxn-py: > Let's remove the Python bits and do it in a follow-up PR.

Sure, that sounds good. On it

arjxn-py: Done in [`874b49f`](https://github.com/jupytercad/JupyterCAD/pull/413/commits/874b49f90c2cc4a0b06ebfbd9c16ec6dde7f8e50)

trungleduc (approved):

trungleduc:
we should update https://github.com/jupytercad/JupyterCAD-FreeCAD since it also uses the `guidata` section

arjxn-py:
> we should update https://github.com/jupytercad/JupyterCAD-FreeCAD since it also uses the `guidata` section

Yes, I'll be on that next. Thank You :)

martinRenou (approved):
Thanks! Let's make an alpha release of this so we can start updating jupytercad-freecad

arjxn-py:
> Thanks! Let's make an alpha release of this so we can start updating jupytercad-freecad

Sounds good, should I wait for the release to open a PR on jupytercad-freecad?

martinRenou:
That will be easier for you that way 👍🏽

martinRenou:
It's out https://github.com/jupytercad/JupyterCAD/releases/tag/v3.0.0a0 You can install it with pip: https://pypi.org/project/jupytercad/3.0.0a0/

PR #415, opened 2024-09-13
Exclude node_modules from sdists

martinRenou:


python/jupytercad_core/pyproject.toml
```diff
@@ -42,7 +42,7 @@ fields = ["description", "authors", "urls"]
 
 [tool.hatch.build.targets.sdist]
 artifacts = ["jupytercad_core/labextension"]
-exclude = [".github", "binder"]
+exclude = [".github", "binder", "node_modules"]
```
arjxn-py, line 45: Question: is `binder` here related to the thing - launch the binder that was being used on PRs before?
Are we still using that?

trungleduc:
it looks like the jupyterlite deployment is broken, could you bump the `actions/deploy-pages` action to v4 to see if if fixes the issue?

python/jupytercad_core/pyproject.toml
trungleduc: we are not using it in the PR but it can be useful for others who want to deploy jupytercad to mybinder

PR #417, opened 2024-09-13
Remove unneeded jupytercad_lab dependencies

martinRenou:
This was having the downside of not being able to install it in an `emscripten-wasm32` env:

```
error    libmamba Could not solve for environment specs
    The following packages are incompatible
    └─ jupytercad_lab is not installable because there are no viable options
       ├─ jupytercad_lab [1.0.0|1.0.1] would require
       │  └─ y-py >=0.6,<0.7 , which does not exist (perhaps a missing channel);
       └─ jupytercad_lab [2.0.0|2.0.1|2.0.2] would require
          └─ jupyterlab >=4,<5 , which requires
             └─ tornado >=6.2.0 , which does not exist (perhaps a missing channel).
```

### DIFF
--- a/python/jupytercad_lab/pyproject.toml
+++ b/python/jupytercad_lab/pyproject.toml
@@ -23,10 +23,7 @@ classifiers = [
   "Programming Language :: Python :: 3.12",
 ]
 dependencies = [
-  "jupyter_server>=2.0.1,<3",
-  "jupyter-ydoc>=2,<3",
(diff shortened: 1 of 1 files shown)


arjxn-py (approved):
Thanks @martinRenou, looks alright to me :)

trungleduc (approved):

trungleduc:
Thanks!

Issue #344, opened 2024-03-29
Update ThreeJS


martinRenou:
We should update ThreeJS and ThreeJS typings. This will allow us to get rid of a couple of `@ts-ignore`s in the mainview.

arjxn-py:
Closing since #418 is merged.

martinRenou:
Thanks Arjun!

Issue #406, opened 2024-09-10
Improve the UX of the clip-plane and wireframe buttons


martinRenou:
## Problems

Today, the clip-plane toolbar button shows a dialog to select two options:
- selected
- show Clip Plane

In the other side, the wireframe button does not show a dialog and directly apply the wireframe effect. Though once toggled, there is no visual indication in the toolbar that the button has been toggled.

Also, refreshing the page or reopening the file does not keep the clip-plane or wireframe states. We could fix this by saving the states in the schema.

## Suggestions

- We could make the behavior more homogeneous by making the clip-plane toolbar button not showing a dialog and directly toggling the clip-plane effect (we would always show the clip-plane in the 3D scene by default).
- We should show a visual indication in the toolbar that toggle buttons are toggled
- We should save the states of the clip-plane and wireframe effect in the schema

arjxn-py:
Thanks @martinRenou, it is well defined. I'll be happy to look into it 🚀

PR #418, opened 2024-09-15
Update `ThreeJS`

arjxn-py:
Currently a draft, but open for reviews
Related to #344


arjxn-py:
image
The changes doesn't look very pretty at the moment:
- We've this plane visible which should be there but not be visible afaik
- The colors of the model are a bit dimmed (the set color on the cube is sharp yellow, Maybe a lightning issue?) 

Considering this a start, the next thing I plan are - 
- [ ] to get rid of some of `@ts-ignores` in the `mainview`
- [ ] Update threeJS typings maybe
- [ ] Identify above issues and resolve them

martinRenou:
Thanks for starting this!

I believe the black mesh we are seeing here is the `this._clippingPlaneMesh` mesh object from the 3D view. IIRC it is not supposed to be visible, this one should only be used for the filling the gaps when clipping is enabled. I can have a deeper look at this later if you need.

arjxn-py:
> I believe the black mesh we are seeing here is the `this._clippingPlaneMesh` mesh object from the 3D view. IIRC it is not supposed to be visible, this one should only be used for the filling the gaps when clipping is enabled. I can have a deeper look at this later if you need.

yes, i've seen this plane in the mainview. I shall try to look into it and also see if it can be fixed by making it not visible or something. I'll be happy to follow up on the same in case I would need you to have to take a look into the same :)

packages/base/src/3dview/mainview.tsx
```diff
@@ -668,28 +668,6 @@ export class MainView extends React.Component<IProps, IStates> {
...
-      stencilZPass: THREE.ReplaceStencilOp,
-      side: THREE.DoubleSide,
-      wireframe: this.state.wireframe
-    });
-    this._clippingPlaneMesh = new THREE.Mesh(planeGeom, planeMat);
-    this._clippingPlaneMesh.onAfterRender = renderer => {
-      renderer.clearStencil();
-    };
-
-    this._scene.add(this._clippingPlaneMesh);
```
arjxn-py, line 692: Hi @martinRenou, removing this plane removes the default plane that's being rendered in the scene. I don't see any issue arised after removing this (especially in clipping), Are we good to remove this?

martinRenou: Hi! 

That plane is used for filling the shapes cut by the clip plane:

Screenshot from 2024-09-24 10-46-59

If we don't show it, the object will show as being non-solid once cut. So I believe we can't get rid of this code. I wonder if we should do `this._clippingPlaneMesh.visible = false` when clipping is not enabled, true otherwise.

arjxn-py:
Seems like I first want to (Improve the UX of the clip-plane and wireframe buttons #406) as I'm trying to update scene everytime clipping button is clicked to update the visiblity of the clip plane render thing.

packages/base/src/3dview/mainview.tsx
arjxn-py: Thanks @martinRenou, I tried implementing your suggestion and now the _clipPlaneMesh is not visible by default at least but it doesn't look enough as in our case when we've clipping enabled we still do not have the whole clipPlaneMesh visible but it just covers the intersection of the mesh with clipPlane. 
But it's fully visible now: 
image

I'm on it trying debugging it but I believe your suggestion would accelerate the process :)

arjxn-py: It took me some time to figure out but was only a one line fix.

packages/base/src/3dview/mainview.tsx
```diff
@@ -218,7 +218,8 @@ export class MainView extends React.Component<IProps, IStates> {
 
       this._renderer = new THREE.WebGLRenderer({
         alpha: true,
-        antialias: true
+        antialias: true,
+        stencil: true
```
martinRenou, line 222: 🚀 Neat

packages/base/package.json
```diff
@@ -60,12 +60,11 @@
     "@naisutech/react-tree": "^3.0.1",
     "@rjsf/core": "^4.2.0",
     "@types/d3-color": "^3.1.0",
-    "@types/three": "^0.134.0",
```
martinRenou, line 63: Did you mean to remove this? Have you tried updating it to `0.168.0`?

arjxn-py: No i didn't mean to remove it, when i did `npm install @types/three` it removed by itself and was also not giving any failure.

arjxn-py: > Have you tried updating it to `0.168.0`?

I'm on it, it adds back the dependency, pushing in a moment.

arjxn-py: Now it fails xD 🫠

arjxn-py: Done in [`b1c311a`](https://github.com/jupytercad/JupyterCAD/pull/418/commits/b1c311a9171efd4c8d3c9f0cc4c47b3e469ba138)

arjxn-py:
Bot please update snapshots.

martinRenou:
Bot please update snapshots!

packages/base/src/3dview/mainview.tsx
```diff
@@ -685,6 +686,7 @@ export class MainView extends React.Component<IProps, IStates> {
       wireframe: this.state.wireframe
     });
     this._clippingPlaneMesh = new THREE.Mesh(planeGeom, planeMat);
+    this._clippingPlaneMesh.visible = this._clipSettings.enabled;
```
martinRenou, line 689: Do we still need this now that we set `stencil: true` on the renderer?

I'm fine letting this here though, it does sound sensible to not render the plane in anyway when the clipping is disabled.

arjxn-py:
> Bot please update snapshots!

Update Galata References workflow is failing everytime xref: #438, I'll be happy to try on to fix it

martinRenou (commented):
Thanks a lot! That looks good to me and seems to work well

packages/base/src/3dview/mainview.tsx
martinRenou: ^^ Just a question, not a blocker

arjxn-py: > Do we still need this now that we set `stencil: true` on the renderer?

Yes, it's necessary to be there as it avoids the clip plane mesh from rendering itself fully and renders only the part of the plane which intersect with the geometry. 

It avoids this from happening: 

image

martinRenou: Ok then let's keep it! I'm unsure what changed in ThreeJS that requires this change, but the code looks more sensible now so that's good I guess

martinRenou:
> Update Galata References workflow is failing everytime xref: https://github.com/jupytercad/JupyterCAD/issues/438, I'll be happy to try on to fix it

I commented there. Would you be able to try and fix the bot so we can get this PR green?

martinRenou:
Bot please update snapshots!!

martinRenou:
Neat! I'll trigger the CI

martinRenou (approved):
Thanks!

PR #422, opened 2024-09-16
Fix python api

trungleduc:


python/jupytercad_lab/jupytercad_lab/notebook/cad_document.py
```diff
@@ -166,32 +165,6 @@ def remove_annotation(self, annotation_id: str) -> None:
             with self.ydoc.transaction() as t:
                 self._metadata.pop(t, annotation_id, None)
 
-    def set_color(self, object_name: str, color: Optional[List]) -> None:
```
trungleduc, line 169: We need to re-add this method using the new `color` field of the shared model

martinRenou: Ok let's do that separately. This should be resolved before final 3.0.0

martinRenou (approved):
Thanks

PR #435, opened 2024-09-19
WIP: Add ViewHelper

arjxn-py:


martinRenou:
I'm excited about this! 😄 

This may be an opportunity for getting rid of the Axis helper thingy, as I believe we won't need it with these changes?

arjxn-py:
> I'm excited about this! 😄

So am I, and as I mentioned it was not looking good before as it was messing with the scene by rendering it skewed. After good amount of time spent it's atleast looking better than that but still not in the best shape to be ready, there are a couple of issues which I'm wondering, why?


https://github.com/user-attachments/assets/502137ce-155a-4030-8965-7d7f8b57679f

- There's this strange offset to the viewHelper you can notice while i hover over it & inspite of me giving it `top-right` position, the div container of viewHelper is in top right but viewHelper in the scene is not.
image

- ViewHelper is not visible when left panel is on and even if it's off it's still not fully visible maybe due to still some space in the left?



> This may be an opportunity for getting rid of the Axis helper thingy, as I believe we won't need it with these changes?

Yes I agree, this can be a good option for axis helper too. :)

arjxn-py:
Superseded by #440

PR #440, opened 2024-10-01
Add viewHelper v2

### LINKED ISSUE #419: Add `ViewHelper` to JupyterCAD Editor
This issue proposes the addition of a `ViewHelper` to the JupyterCAD editor. This would greatly enhance user experience by providing a 3D viewport gizmo that allows users to interactively control and view camera orientations.

The `ViewHelper` can offer users a visual and interactive tool to change views, similar to the feature demonstrated in this repository: [three-viewport-gizmo](https://github.com/Fennec-hub/three-viewport-gizmo). A working example can be found here: [three-viewport-gizmo example](https://fennec-hub.github.io/three-viewport-gizmo/).

### Expected Behavior

- A small interactive sphere-like gizmo should appear in the top or bottom-right corner of the JupyterCAD canvas.
- Users should be able to click on faces, edges, and corners of the gizmo to change the view orientation (top, bottom, left, right, front, back).
- The `ViewHelper` should remain fixed on the screen and should not be affected by camera movements such as panning, zooming, or rotating the 3D model in the main scene.
- It should provide smooth transitions between camera views, enhancing navigation in complex 3D scenes.

### Proposed Implementation

- The `three-viewport-gizmo` library can be utilized to integrate the `ViewHelper` functionality within the JupyterCAD editor.
- The gizmo should be rendered separately from the main 3D scene to avoid interference with the model itself, and it should be responsive to user interactions (clicks and pointer movements).
- See the example on how this can be implemented: [three-viewport-gizmo example](https://fennec-hub.github.io/three-viewport-gizmo/).

arjxn-py:
https://github.com/user-attachments/assets/7d633f26-28b6-49aa-bb53-5370997e9a7c

Should close #419


packages/base/src/3dview/mainview.tsx
```diff
@@ -377,6 +399,7 @@ export class MainView extends React.Component<IProps, IStates> {
     this._renderer.setRenderTarget(null);
     this._renderer.clearDepth();
     this._renderer.render(this._scene, this._camera);
+    this._viewHelper.render(this._renderer);
```
martinRenou, line 402: What happens if you add the viewhelper to the scene instead of calling this? (that way it gets renderer with the scene)

arjxn-py: This way viewHelper gets added to the scene and it's no longer independent of the scene panning, plus by default it gets added to the center of the scene like this - 
image

packages/base/src/3dview/mainview.tsx
```diff
@@ -377,6 +399,8 @@ export class MainView extends React.Component<IProps, IStates> {
     this._renderer.setRenderTarget(null);
     this._renderer.clearDepth();
     this._renderer.render(this._scene, this._camera);
+    this._renderer.autoClear = false;
+    this._viewHelper.render(this._renderer);
```
martinRenou, line 403: We should set `autoClear` to false only once when creating the renderer, in here though, we need to call `clear` and not only `clearDepth` to both clear the color and depth buffers

```suggestion
    this._renderer.clear();
    this._renderer.render(this._scene, this._camera);
    this._viewHelper.render(this._renderer);
```

arjxn-py: Thanks, merged in [`ffa3a5b`](https://github.com/jupytercad/JupyterCAD/pull/440/commits/ffa3a5bb701eb355685a2c6532669e72916df89f)

arjxn-py:
bot please update snapshots

packages/base/src/3dview/mainview.tsx
```diff
@@ -372,11 +397,15 @@ export class MainView extends React.Component<IProps, IStates> {
...
       );
     }
+    if (this._viewHelper.animating) {
+      this._viewHelper.update(this._delta);
+    }
 
     this._controls.update();
     this._renderer.setRenderTarget(null);
-    this._renderer.clearDepth();
+    this._renderer.autoClear = false;
```
martinRenou, line 406: What I meant in https://github.com/jupytercad/JupyterCAD/pull/440#discussion_r1782971785 

was that this line should be:

```suggestion
    this._renderer.clear();
```

And that `autoClear` should be set to false only upon creation of the renderer (no need to set it on every frame). So maybe adding `this._renderer.autoClear = false;` around line 228 from this file

arjxn-py: Thanks @martinRenou, done in [`5d8dc84`](https://github.com/jupytercad/JupyterCAD/pull/440/commits/5d8dc84cfc7ce7e02be3783b75744403efabecfb)

arjxn-py:
bot please update snapshots

arjxn-py:
bot please update snapshots

packages/base/src/3dview/mainview.tsx
```diff
@@ -1411,6 +1441,9 @@ export class MainView extends React.Component<IProps, IStates> {
   private _controls: OrbitControls; // Mouse controls
   private _transformControls: TransformControls; // Mesh position/rotation controls
   private _pointer3D: IPointer | null = null;
+  private _clock: THREE.Clock;
+  private _delta: number;
```
trungleduc, line 1445: Do we need `_delta` as a class property?

arjxn-py: I just confirmed and it's not needed as a class property.
This has been added as class property in the first place as when i defined it as const the first time it was always 0 for some reason and when i gave it let it was giving error as it's never reassigned, then i did this. On my way! to define it as animate method property now.

arjxn-py: Done in [`2923de0`](https://github.com/jupytercad/JupyterCAD/pull/440/commits/2923de004316aa28c93747efddc83d3b2c460546)

martinRenou (approved):
This is neat! Thanks

Let's remove our legacy axes helper from the codebase in a separate PR

PR #441, opened 2024-10-02
Improve the UX of the clip-plane and wireframe buttons

### LINKED ISSUE #406: Improve the UX of the clip-plane and wireframe buttons

arjxn-py:
https://github.com/user-attachments/assets/42d2a6d4-728a-49d0-a914-c45b449bffe4

Should close #406


arjxn-py:
Saving these properties in schema to preserve state remains.

packages/base/src/commands.ts
```diff
@@ -971,6 +937,17 @@ export function addCommands(
         return;
       }
       current.wireframe = !current.wireframe;
+
+      const buttonNode = document.querySelector(
+        `[data-command="${CommandIDs.wireframe}"]`
+      );
```
trungleduc, line 943: If I have multiple CAD documents opened. This selector will not select the correct one.

packages/base/src/commands.ts
```diff
@@ -1079,15 +1056,30 @@ export function addCommands(
...
+
+      const { enabled, showClipPlane } = panel.clipView;
+      panel.clipView = {
+        enabled: enabled,
+        showClipPlane: showClipPlane
+      };
+
+      const buttonNode = document.querySelector(
+        `[data-command="${CommandIDs.updateClipView}"]`
+      );
```
trungleduc, line 1075: Same here

packages/base/src/commands.ts
arjxn-py: Ahh yes, you're right

martinRenou (approved):
Thanks!

PR #447, opened 2024-10-05
Toggle buttons working correctly when multiple editors are open

arjxn-py:


packages/base/src/commands.ts
```diff
@@ -942,6 +942,9 @@ export function addCommands(
       }
       current.wireframe = !current.wireframe;
       commands.notifyCommandChanged(CommandIDs.wireframe);
+      tracker.currentChanged.connect(() => {
+        commands.notifyCommandChanged(CommandIDs.wireframe);
+      });
```
martinRenou, line 947: I believe you need to extract this logic outside of the command's `execute`. Otherwise you will not listen to `tracker.currentChanged` unless you already executed the command.

arjxn-py: Did in [`7a1954b`](https://github.com/jupytercad/JupyterCAD/pull/447/commits/7a1954ba67ceca4c7314f994ba73d25a9179a66d), have i done it right?

martinRenou: That's looks ok to me! Thanks!

martinRenou (approved):
🚀

PR #448, opened 2024-10-05
Sync editor upon editor change

### LINKED ISSUE #450: Creating new `.jcad` file while having another open makes the scene blank for existing one
https://github.com/user-attachments/assets/aca9f859-7058-4562-be91-153bf6617a2f

arjxn-py:
Should close #450


martinRenou (commented):
Neat!

I don't quite like that this lives under `commands.ts` though. I'd suggest we put this code somewhere in the `JupyterCadPanel` class (under `packages/base/src/widget.ts`)

trungleduc:
I agree with @martinRenou

martinRenou:
With your recent changes you are still calling this logic in `commands.ts` which should not responsible for this IMO. I'd suggest doing this **in** the `JupyterCadPanel` class implementation, maybe in the constructor.

This may be a bit more involved as you'd need for the `tracker` to be an option for the `JupyterCadPanel` constructor, but I believe it's the way to go.

This way the JupyterCadPanel is responsible for resizing itself when the context changes. `commands.ts` is merely a place to define our commands.

arjxn-py:
Thanks for the comprehensive explanation Martin. This sounds good, i'll be on it.

packages/base/src/widget.ts
```diff
@@ -67,6 +69,17 @@ export class JupyterCadPanel extends SplitPanel {
     SplitPanel.setStretch(this._jupyterCadMainViewPanel, 1);
   }
 
+  private _syncEditor(tracker: WidgetTracker<IJupyterCadWidget>) {
```
martinRenou, line 72: Nitpicking suggestion:

I would not define a method for this as this is called just once in the constructor. So maybe just inline that code there :)

martinRenou (commented):
Neat! I just have a final suggestion

packages/base/src/widget.ts
martinRenou: Ah right! I agree. Sorry Arjun for the misleading comments earlier

arjxn-py: No worries, It's been a good mental exercise 🤣 
I'll be on it.

arjxn-py: Done in [`4115b79`](https://github.com/jupytercad/JupyterCAD/pull/448/commits/4115b79a5cb1a5d636ed4982df147d465364c8c4)
And I believe I myself am responsible that this PR took longer than usual as my initial iterations of solution were not close to perfect.

trungleduc (approved):

trungleduc:
Thanks!

SylvainCorlay:
Ah weird, the merge commit does not pass the integration tests.

arjxn-py:
> Ah weird, the merge commit does not pass the integration tests.

Yes, should be related to #471

PR #452, opened 2024-10-07
Button to Center view on visible objects

### LINKED ISSUE #127: Button to center view on visible objects
After some operations, the view is quickly not centered anymore on the object. We should have a button that re-center the view to the visible objects on-click.

arjxn-py:
Should close #127 

https://github.com/user-attachments/assets/8f478f76-6c2c-4aed-b527-b5dd641889d8

### DIFF
--- a/packages/base/src/3dview/mainview.tsx
+++ b/packages/base/src/3dview/mainview.tsx
@@ -120,7 +120,7 @@ export class MainView extends React.Component<IProps, IStates> {
     window.addEventListener('jupytercadObjectSelection', (e: Event) => {
       const customEvent = e as CustomEvent;
 
-      if (customEvent.detail.mainViewModel === this._mainViewModel.id) {
+      if (customEvent.detail.mainViewModelId === this._mainViewModel.id) {
         this.lookAtPosition(customEvent.detail.objPosition);
       }
     });
--- a/packages/base/src/3dview/mainview.tsx
+++ b/packages/base/src/3dview/mainview.tsx
@@ -119,7 +119,10 @@ export class MainView extends React.Component<IProps, IStates> {
     this._mainViewModel.initSignal();
     window.addEventListener('nodeClick', (e: Event) => {
       const customEvent = e as CustomEvent;
-      this.lookAtPosition(customEvent.detail.objPosition);
+
+      if (customEvent.detail.mainViewModel === this._mainViewModel.id) {
+        this.lookAtPosition(customEvent.detail.objPosition);
+      }
     });
   }
 
--- a/packages/base/src/3dview/mainview.tsx
+++ b/packages/base/src/3dview/mainview.tsx
@@ -117,6 +117,10 @@ export class MainView extends React.Component<IProps, IStates> {
     this.addContextMenu();
     this._mainViewModel.initWorker();
     this._mainViewModel.initSignal();
+    window.addEventListener('nodeClick', (e: Event) => {
+      const customEvent = e as CustomEvent;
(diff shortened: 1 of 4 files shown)


packages/base/src/3dview/mainview.tsx
```diff
@@ -117,6 +117,10 @@ export class MainView extends React.Component<IProps, IStates> {
     this.addContextMenu();
     this._mainViewModel.initWorker();
     this._mainViewModel.initSignal();
+    window.addEventListener('nodeClick', (e: Event) => {
```
trungleduc, line 120: This will also listen to events coming from other CAD documents.

packages/base/src/panelview/objecttree.tsx
```diff
@@ -237,6 +237,26 @@ class ObjectTreeReact extends React.Component<IProps, IStates> {
...
+    const object = this.getObjectFromName(objectId);
+
+    if (object && object.visible === true) {
+      const objPosition = object?.parameters?.Placement?.Position || {
+        x: 0,
+        y: 0,
+        z: 0
+      };
+
+      const event = new CustomEvent('nodeClick', {
```
trungleduc, line 250: you need to add the `MainViewModel.id` to this event to be able to identify the correct 3D view

arjxn-py: Thanks @trungleduc, for that I'd need ho access `MainViewModel.id` in objectTree component to emit it correctly. I'll be on it and update :)

trungleduc: Something like this in the `ControlPanelModel`:

```typescript
  get mainViewModel(): string | undefined {
    return (this._tracker.currentWidget as JupyterCadWidget | null)?.content
      .currentViewModel.id;
  }
```
Then in `ObjectTreeReact` you can use `this.props.cpModel.mainViewModel?.id` to get the current main view id

trungleduc (changes requested):

packages/base/src/panelview/objecttree.tsx
arjxn-py: Super @trungleduc, this helps so much 🚀

martinRenou:
Neat! Can you show a screencast of the feature?

arjxn-py:
> Neat! Can you show a screencast of the feature?

Sure:

https://github.com/user-attachments/assets/8f478f76-6c2c-4aed-b527-b5dd641889d8

packages/base/src/3dview/mainview.tsx
martinRenou: Adding my 2 cents, what Trung means here is that if you have two docs opened, the two mains views will update when clicking in the left panel. This is not really what you want.

https://github.com/user-attachments/assets/28fe02ef-ded0-4a0f-93e6-f73b9057781c

packages/base/src/panelview/objecttree.tsx
```diff
@@ -237,6 +237,27 @@ class ObjectTreeReact extends React.Component<IProps, IStates> {
...
```
martinRenou, line 250: I did not know one could define a custom event this way!

I guess we should be a tiny bit more specific about the event name. Maybe an event name like `jupytercadObjectSelection` would be better?

arjxn-py: Done! Thanks for suggesting.

martinRenou (approved):
Thanks!

arjxn-py:
Looks like this PR is angry and requires review from @trungleduc again since he requested changes?

PR #455, opened 2024-10-08
Reactive toolbar w.r.t. width

### LINKED ISSUE #445: Toolbar Not Fully Accessible When JupyterCad Editor Width Is Reduced
When the JupyterCad editor's width is reduced beyond a certain point, the toolbar becomes partially inaccessible, making it difficult to use essential buttons. This issue impacts user experience, particularly when working in smaller screen resolutions or when the editor is resized.

arjxn-py:
Should close #445 
Thanks to @brichet

image


arjxn-py:
Bot please update snapshots.

martinRenou (approved):
Thanks!

Let's port this to jupytergis too!

martinRenou:
Please update snapshots

arjxn-py:
> Let's port this to jupytergis too!

Sounds Good, happy to do that.

martinRenou:
The visual regression test failure looks legit. It seems our console toggle test fails on:

```typescript
await page.getByRole('button', { name: 'Toggle console' }).click();
```

Probably the selector by role is broken now with the new toolbar?

arjxn-py:
> Probably the selector by role is broken now with the new toolbar?

Helpful hint, I have to look into it

arjxn-py:
It might be because https://github.com/jupytercad/JupyterCAD/blob/main/ui-tests/tests/ui.spec.ts-snapshots/JCAD-Console-linux.png has less width?

And Toggle console button is not accessible without clicking on three dots in that case.

Screenshot 2024-10-09 at 17 43 47

arjxn-py:
Oh, i think test need to be fixed in another PR and has to be in `main` before this one?

ui-tests/tests/ui.spec.ts
```diff
@@ -358,7 +358,11 @@ test.describe('UI Test', () => {
   test.describe('Console activation test', () => {
     test('should open console', async ({ page }) => {
       await page.goto();
-      await page.getByLabel('notebook content').getByText('CAD File').click();
+      await page
+        .getByLabel('notebook content')
+        .getByText('New JCAD File')
```
martinRenou, line 363:
```suggestion
        .getByText('CAD File')
```

martinRenou:
Triggering the CI

PR #466, opened 2024-10-10
Add `logoicon` as tabicon

### LINKED ISSUE #459: Missing icon in dock area tab header
We are missing the icon in the tab header

Image

This is probably related to #456.

arjxn-py:
Should close #459 

image


arjxn-py:
> We should do the same for jupytergis

Done https://github.com/geojupyter/jupytergis/pull/183 😁

PR #469, opened 2024-10-11
Fix breaking viewHelper while changing camera

### LINKED ISSUE #463: Changing the camera projection type breaks the view helper
Changing the camera projection seems to break the view helper

https://github.com/user-attachments/assets/02288e03-90f3-41c3-9bc2-83af7f401105

arjxn-py:
Fixes #463 

https://github.com/user-attachments/assets/443497ee-5ce9-4731-ae5a-4f62bee59d0e

### DIFF
--- a/packages/base/src/3dview/mainview.tsx
+++ b/packages/base/src/3dview/mainview.tsx
@@ -362,29 +362,29 @@ export class MainView extends React.Component<IProps, IStates> {
 
       this._transformControls.enabled = false;
       this._transformControls.visible = false;
-
-      // ViewHelper setup
-      this._viewHelper = new ViewHelper(
-        this._camera,
-        this._renderer.domElement
-      );
-      this._viewHelper.center = this._controls.target;
-
-      const viewHelperDiv = document.createElement('div');
-      viewHelperDiv.id = 'viewHelper';
-      viewHelperDiv.style.position = 'absolute';
-      viewHelperDiv.style.right = '0px';
-      viewHelperDiv.style.bottom = '0px';
-      viewHelperDiv.style.height = '128px';
-      viewHelperDiv.style.width = '128px';
-
-      this.divRef.current.appendChild(viewHelperDiv);
-
-      viewHelperDiv.addEventListener('pointerup', event =>
-        this._viewHelper.handleClick(event)
-      );
+      this._createViewHelper();
     }
   };
+
+  private _createViewHelper() {
+    this._viewHelper = new ViewHelper(this._camera, this._renderer.domElement);
+    this._viewHelper.center = this._controls.target;
+
+    const viewHelperDiv = document.createElement('div');
+    viewHelperDiv.style.position = 'absolute';
+    viewHelperDiv.style.right = '0px';
+    viewHelperDiv.style.bottom = '0px';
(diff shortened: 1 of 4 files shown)


packages/base/src/3dview/mainview.tsx
```diff
@@ -362,29 +362,33 @@ export class MainView extends React.Component<IProps, IStates> {
...
-      this._viewHelper = new ViewHelper(
-        this._camera,
-        this._renderer.domElement
-      );
-      this._viewHelper.center = this._controls.target;
+  private _createViewHelper() {
+    const viewHelperDiv = document.getElementById('viewHelper');
+    if (viewHelperDiv) {
+      viewHelperDiv.remove();
+    }
```
trungleduc, line 373: This will remove the view helper of all opening documents. This is the expected behavior?

martinRenou: Yeah, we can probably find a solution here that does not require using an `id` whatsoever, just `this.something.remove()`, something being the view helper container kept as a class property?

arjxn-py: ThankYou both, I agree this was not the expected behaviour even if it was not noticeable while testing for multiple documents open.

I also realised that it was not necessary to remove viewHelperDiv and just updating it with `updateCamera` works as well.


Done in [`39f6648`](https://github.com/jupytercad/JupyterCAD/pull/469/commits/39f664819e084b58636c54d0a203ad4570cf3820)

packages/base/src/3dview/mainview.tsx
```diff
@@ -362,29 +362,29 @@ export class MainView extends React.Component<IProps, IStates> {
...
 
-      this.divRef.current.appendChild(viewHelperDiv);
+    this._viewHelperDiv = viewHelperDiv;
 
-      viewHelperDiv.addEventListener('pointerup', event =>
-        this._viewHelper.handleClick(event)
-      );
-    }
-  };
+    this.divRef.current?.appendChild(this._viewHelperDiv);
```
martinRenou, line 382: Just a question, how are we removing the old one now?

arjxn-py: We're not removing it now as it was not necessary, just updating it with `updateCamera` works.

martinRenou: What I mean is that you re-create a new `this._viewHelperDiv` in `updateCamera`, that you append to `this.divRef.current`.

Don't we need to remove the old `this._viewHelperDiv` that was created before?

arjxn-py: image

Yes you're right, I just verified. Sorry that I overlooked, I'll be omw to fix it.

arjxn-py: Done in [`4602e00`](https://github.com/jupytercad/JupyterCAD/pull/469/commits/4602e00cff6d62640a924ed800558c256c3c22fe)

Thank You for pointing this out :)

martinRenou (approved):
Thanks ! 🚀 

Failed UI tests are not related https://github.com/jupytercad/JupyterCAD/issues/471

PR #490, opened 2024-10-16
Exaggerate clip plane size

martinRenou:
The clipplane mesh is not fully filling the gap left by the clipping effect. This PR fixes it.

Before:

https://github.com/user-attachments/assets/b59b2d6e-d6bc-476c-b3a0-195e2bcf23f7

After:

https://github.com/user-attachments/assets/787da56a-b826-450c-8bdb-8854ac895824

### DIFF
--- a/packages/base/src/3dview/mainview.tsx
+++ b/packages/base/src/3dview/mainview.tsx
@@ -747,8 +747,8 @@ export class MainView extends React.Component<IProps, IStates> {
 
     // Clip plane rendering
     const planeGeom = new THREE.PlaneGeometry(
-      this._refLength! * 10, // *10 is a bit arbitrary and extreme but that does not impact performance or anything
-      this._refLength! * 10
+      this._refLength! * 1000, // *1000 is a bit arbitrary and extreme but that does not impact performance or anything
+      this._refLength! * 1000
     );
     const planeMat = new THREE.MeshPhongMaterial({
       color: DEFAULT_EDGE_COLOR,


trungleduc (approved):

arjxn-py (approved):
Super 🚀 
Thanks @martinRenou

SylvainCorlay:
Could this be computed base on the e.g. the bounding box of the whole jcad document, or would that be too expensive? (I guess bounding boxes of individual objects could be cached)

trungleduc:
> Could this be computed base on the e.g. the bounding box of the whole jcad document, or would that be too expensive? (I guess bounding boxes of individual objects could be cached)

`this._refLength` is the size of the bounding box of the whole jcad document.

martinRenou:
> How do you turn these controls to the sphere thing? That's soo coool

That's the "R" keyboard shortcut :)

martinRenou:
> this._refLength is the size of the bounding box of the whole jcad document.

By the way, testing out this morning, I see that `refLength` is quite small and is not updating properly.

I'm looking into it. 

Multiplying by 1000 here still makes sense, we just want to make sure the plane is bigger, if I could make an infinite plane I would.

SylvainCorlay:
> By the way, testing out this morning, I see that refLength is quite small and is not updating properly.

Is this still required after https://github.com/jupytercad/JupyterCAD/pull/493?

martinRenou:
I'm not sure why you're saying it might not be good?

martinRenou:
What we need is an infinite plane, but there is no such thing in WebGL. Multiplying the reflength by a big number allows us to make sure the size of the plane is big enough to cover the shapes.

PR #492, opened 2024-10-17
Handle zoom while camera change

### LINKED ISSUE #483: Switching from perspective projection to orthogonal projection zooms out a lot
Switching from perspective projection to orthogonal projection zooms out a lot. We should try to improve this.

https://github.com/user-attachments/assets/d0ac649d-6686-4298-a98a-aa188e3660ad

arjxn-py:
https://github.com/user-attachments/assets/edb44fbc-1fe9-401e-9156-63a3aa6a10c3

Shall fix #483

### DIFF
--- a/packages/base/src/3dview/mainview.tsx
+++ b/packages/base/src/3dview/mainview.tsx
@@ -1313,6 +1313,7 @@ export class MainView extends React.Component<IProps, IStates> {
   private _updateCamera() {
     const position = new THREE.Vector3().copy(this._camera.position);
     const up = new THREE.Vector3().copy(this._camera.up);
+    const target = this._controls.target.clone();
 
     this._camera.remove(this._cameraLight);
     this._scene.remove(this._camera);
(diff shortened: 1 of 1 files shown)


packages/base/src/3dview/mainview.tsx
```diff
@@ -1323,12 +1324,17 @@ export class MainView extends React.Component<IProps, IStates> {
       const width = this.divRef.current?.clientWidth || 0;
       const height = this.divRef.current?.clientHeight || 0;
 
+      const distance = position.distanceTo(target);
+      const zoomFactor = 1000 / distance;
```
martinRenou, line 1328: Do you know why we need this magic `1000` number?

arjxn-py: This is kind of an arbitrary number after some hit and trials, I noticed that more you zoom the object in `PerspectiveCamera` more further it goes on `OrthographicCamera`

arjxn-py: So it meant that the distance was inversely proportional to the zooming in that i needed to do

martinRenou: Sounds good! I tested on my side and this seems to work well. We have other magic numbers in JupyterCAD

PR #495, opened 2024-10-17
Use logarithmic depth buffer

martinRenou:
This prevents visual artifacts with big/small objects


arjxn-py:
Bot please update snapshots

ui-tests/tests/ui.spec.ts-snapshots/JCAD-Undo-linux.png
martinRenou, line 1: 👀 wut

arjxn-py (approved):
Works well, thanks @martinRenou 

The failure looks unrelated, maybe you want to re-trigger that one?

martinRenou:
The visual regression makes me nervous though, I don't understand why the color of the box changes in that case

arjxn-py:
> The visual regression makes me nervous though, I don't understand why the color of the box changes in that case

It must be since the `SELECTED_MESH_COLOR` is removed, not sure why it's no longer selected with bounding box and why it's happening in this PR in the first place

martinRenou:
Ok well let's go and merge! Thanks

PR #497, opened 2024-10-17
Bump pydantic==2.9.2 for build

davidbrochart:
This makes it possible to build JupyterCAD with Python 3.13 without a Rust compiler.


martinRenou (approved):
Thanks!

PR #498, opened 2024-10-17
Streamline edge selection

### LINKED ISSUE #478: Selecting edges used to be easier
3D meshes used to have "sticky" edges, I recall this was achieved by having edges have a larger hitbox than their actual visual width.

It looks like it is not the case anymore (not sure since when?).

The result of this is that it is now more difficult to select edges as the hit needs to be pixel perfect.

arjxn-py:
Shall fix #478 & #482 

https://github.com/user-attachments/assets/b9985faf-5095-43e1-8a63-1786cb84925a

### DIFF
--- a/packages/base/src/3dview/mainview.tsx
+++ b/packages/base/src/3dview/mainview.tsx
@@ -103,7 +103,7 @@ export class MainView extends React.Component<IProps, IStates> {
     this._mainViewModel.renderSignal.connect(this._requestRender, this);
     this._mainViewModel.workerBusy.connect(this._workerBusyHandler, this);
 
-    this._raycaster.params.Line = { threshold: 50 };
+    this._raycaster.params.Line2 = { threshold: 50 };
 
     this.state = {
       id: this._mainViewModel.id,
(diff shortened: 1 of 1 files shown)


martinRenou:
This makes me think we should probably have some UI tests about selecting edges and applying edge operations

arjxn-py:
> This makes me think we should probably have some UI tests about selecting edges and applying edge operations. Let's do that in a separate PR!

Yes, that sounds good.

packages/base/src/3dview/mainview.tsx
```diff
@@ -103,7 +103,7 @@ export class MainView extends React.Component<IProps, IStates> {
     this._mainViewModel.renderSignal.connect(this._requestRender, this);
     this._mainViewModel.workerBusy.connect(this._workerBusyHandler, this);
 
-    this._raycaster.params.Line = { threshold: 50 };
+    this._raycaster.params.Line2 = { threshold: 50 };
```
martinRenou, line 106: I am a bit confused by this.

- I see it working properly in Firefox, just like in your screencast. So far so good, but we're not using `Line2` in the codebase anymore, so why does it work?
- I do not see it working in Chromium for some reason.

So should we explicitly use `Line2` instead of `Line` for line meshes in the codebase so that it works in all cases? I would try that, but I really don't understand why it works in Firefox.

martinRenou (approved):
Thanks! Let's merge and iterate. I'll see if I can make it work on my chromium locally.

PR #499, opened 2024-10-17
Fix README comment about freecad

martinRenou:


arjxn-py (approved):
Looks Good, thanks :)

PR #501, opened 2024-10-18
Comparable Default Dimensions

### LINKED ISSUE #486: Default primitive dimensions should be comparable
The default diameter of a sphere is 10, while the default dimensions of a box is 1x1x1.

I think that we should use comparable size for primitives.

Image

arjxn-py:
Screenshot 2024-10-18 at 15 22 47

Shall close #486


martinRenou (approved):
Thanks!

Issue #500, opened 2024-10-17
Clip plane color is not visible on dark background


SylvainCorlay:
It seems that the clip plane color is not quite visible on dark backgrounds. It is still somewhat visible, so it is probably just a color choice issue.

![Image](https://github.com/user-attachments/assets/eb22e536-78d1-4c10-92e9-f694ee7d89d4)

martinRenou:
Adding it to the 3.0.0 milestone.

Right now the clip plane is `color: 'black', opacity: 0.2`. Maybe the color should come from a CSS variable instead, and react on theme change.

PR #503, opened 2024-10-18
Some clip plane improvements

martinRenou:
Explore fixing #500 + explore fixing z-fighting using polygon offset. 

I can't seem to implement the later using the logarithmic depth buffer, so opening as draft for now.

https://github.com/user-attachments/assets/aef66c27-0bec-4e5c-b275-ce086f6552cf

### DIFF
--- a/packages/base/src/3dview/mainview.tsx
+++ b/packages/base/src/3dview/mainview.tsx
@@ -236,7 +236,7 @@ export class MainView extends React.Component<IProps, IStates> {
       this._renderer = new THREE.WebGLRenderer({
         alpha: true,
         antialias: true,
-        stencil: true,
+        stencil: true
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
Maybe the simpler color theme fix could be merged independently first.

martinRenou:
This may be a bit more involved as this will require hacking around the fragment shader to manually offset the depth computation after the logarithmic depth buffer was performed. See #509

SylvainCorlay:
I think that the clip plane remains a case in which solving the z fighting issue makes a lot of sense.

martinRenou:
Definitely

martinRenou:
Replaced by https://github.com/jupytercad/JupyterCAD/pull/539

PR #505, opened 2024-10-20
Fix handling of theme change

### LINKED ISSUE #480: Changing theme in `jupyter cad` makes the scene blank
https://github.com/user-attachments/assets/cf27756c-1311-4914-bac9-386b3987351d

arjxn-py:
https://github.com/user-attachments/assets/d24c193a-874e-4d65-8d63-5adf7e30e738

Shall close #480


SylvainCorlay:
A full resize event would trigger a complete reflow of the page. Maybe just doing it on the view would be more economical.

Also, is the resize the only thing that triggers a redraw of the view?

arjxn-py:
> Also, is the resize the only thing that triggers a redraw of the view?

That sounds good, thanks @SylvainCorlay.
I at the moment can't think of any other way but I'm happy to discuss and iterate over it.

SylvainCorlay:
Just calling `update()` on the Lumino panel of the 3-D view should do it. People more familiar with Lumino on the team will know the proper way to do.

trungleduc:
you can use `MessageLoop.sendMessage` to send the `Widget.ResizeMessage` message to the 3D view widget. 

https://github.com/jupytercad/JupyterCAD/blob/7e6b49b8be3948b9b0dd24aa2db6fc4414ea7f0c/packages/base/src/panelview/formbuilder.tsx#L37

arjxn-py:
> you can use `MessageLoop.sendMessage` to send the `Widget.ResizeMessage` message to the 3D view widget.
> 
> https://github.com/jupytercad/JupyterCAD/blob/7e6b49b8be3948b9b0dd24aa2db6fc4414ea7f0c/packages/base/src/panelview/formbuilder.tsx#L37

@trungleduc When i try `MessageLoop.sendMessage` in model.ts it says 
```
Property 'sendMessage' does not exist on type 
'typeof import("/Users/arjunverma/Desktop/Arjun/JupyterCAD/node_modules/@lumino/messaging/types/index")'
```
 but the same works in formbuilder.tsx

martinRenou:
My 2 cents, we used to depend on the `lightTheme` attribute in the React component state to change the background color according to it: https://github.com/jupytercad/JupyterCAD/blob/main/packages/base/src/3dview/mainview.tsx#L1444

This seems to not be used anymore! So we should probably remove `lightTheme` from the state, but we should probably modify this `_handleThemeChange` method there to force a `this.update()` of the react component to trigger a re-render, instead of setting that old unused state attribute.

martinRenou (approved):
Neat! These are the best PRs

PR #508, opened 2024-10-21
Fix clip plane color

martinRenou:
Fix https://github.com/jupytercad/JupyterCAD/issues/500


arjxn-py (approved):
image
Screenshot 2024-10-21 at 13 26 29

Looks Good, thanks @martinRenou

PR #510, opened 2024-10-21
Orthographic camera: fix missing near and far parameters

martinRenou:
This fixes visual artifacts issues when rendering big/small objects with the orthographic camera


SylvainCorlay:
Should one revert some of https://github.com/jupytercad/JupyterCAD/pull/492

martinRenou:
What do you mean?

SylvainCorlay:
>  What do you mean?

Ah, I thought it was related to zoom level.

martinRenou:
Not at all, these are the camera depth check limits

PR #513, opened 2024-10-21
Fix clip plane color

martinRenou:
The initial color was not correct + after update was also not correct


arjxn-py (approved):
image
image

Fassttt 🚀

PR #514, opened 2024-10-21
Fix lighting issue

arjxn-py:
https://github.com/user-attachments/assets/606d56d9-dccd-434b-bb4f-cb53c97290e8

### DIFF
--- a/packages/base/src/3dview/mainview.tsx
+++ b/packages/base/src/3dview/mainview.tsx
@@ -227,7 +227,8 @@ export class MainView extends React.Component<IProps, IStates> {
 
       this._scene.add(new THREE.AmbientLight(0xffffff, 0.5)); // soft white light
 
-      this._cameraLight = new THREE.PointLight(0xffffff, 1);
+      this._cameraLight = new THREE.PointLight(0xffffff, 100);
+      this._cameraLight.decay = 5;
 
       this._camera.add(this._cameraLight);
 
--- a/packages/base/src/3dview/helpers.ts
+++ b/packages/base/src/3dview/helpers.ts
@@ -161,7 +161,6 @@ export function buildShape(options: {
     wireframe: false,
     flatShading: false,
     clippingPlanes,
-    // shininess: 0,
     metalness: 0.5,
     roughness: 0.5
   });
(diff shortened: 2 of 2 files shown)


trungleduc:
what happens if we keep using the `MeshPhongMaterial`?

arjxn-py:
> what happens if we keep using the `MeshPhongMaterial`?

`metalness` & `roughness` properties doesn't seem to be there in `MeshPhongMaterial`

martinRenou:
The standard material uses PBR so it's actually good we're switching to this!

trungleduc:
> > what happens if we keep using the `MeshPhongMaterial`?
> 
> `metalness` & `roughness` properties doesn't seem to be there in `MeshPhongMaterial`

i know, but are they needed if we use zero decay? Could you post a screenshot of using `MeshPhongMaterial`?

arjxn-py:
> > > what happens if we keep using the `MeshPhongMaterial`?
> > 
> > 
> > `metalness` & `roughness` properties doesn't seem to be there in `MeshPhongMaterial`
> 
> i know, but are they needed if we use zero decay? Could you post a screenshot of using `MeshPhongMaterial`?

on it

arjxn-py:
> i know, but are they needed if we use zero decay? Could you post a screenshot of using `MeshPhongMaterial`?


https://github.com/user-attachments/assets/2d73d287-7a78-4022-85bd-3983ceb789c3

@trungleduc

trungleduc:
Thanks, indeed it looks better with the `MeshStandardMaterial`

trungleduc (approved):

packages/base/src/3dview/mainview.tsx
```diff
@@ -227,7 +227,8 @@ export class MainView extends React.Component<IProps, IStates> {
 
       this._scene.add(new THREE.AmbientLight(0xffffff, 0.5)); // soft white light
 
-      this._cameraLight = new THREE.PointLight(0xffffff, 1);
+      this._cameraLight = new THREE.PointLight(0xffffff, 100);
+      this._cameraLight.decay = 5;
```
martinRenou, line 231: I thought we would be using decay `0` with an intensity of `1`?

arjxn-py: Oh Shoot 😬 , this is my bad I think i pushed the wrong changes as I was changing them rapidly. So Sorry

martinRenou: No worries!

PR #516, opened 2024-10-21
Bump @jupyter/ydoc

trungleduc:


trungleduc:
Merging myself to unblock the releaser

PR #521, opened 2024-10-22
Fix bounding box on exploded view

### LINKED ISSUE #491: The bounding box does seems to exagerate the exploded view factor
Image

arjxn-py:
https://github.com/user-attachments/assets/ff24266d-9493-4911-b8cc-90c80fcd68dd

Shall fix #491


packages/base/src/3dview/mainview.tsx
```diff
@@ -1064,6 +1071,7 @@ export class MainView extends React.Component<IProps, IStates> {
         boundingBox.position.copy(center);
 
         selectedMesh.add(boundingBox);
```
martinRenou, line 1073: Shall we not do this anymore? Now that you add it to the group I feel like we should not also add it to the mesh?

arjxn-py: Yes, that sounds good.

packages/base/src/3dview/mainview.tsx
```diff
@@ -1000,6 +1000,13 @@ export class MainView extends React.Component<IProps, IStates> {
         selectedMesh.remove(boundingBox);
       }
```
martinRenou, line 1001: I believe we also don't need to remove `boundingBox` from `selectedMesh` too, now that you don't add it to the mesh anymore :)

martinRenou (commented):
Working great!!!

packages/base/src/3dview/mainview.tsx
arjxn-py: Oops! I should've thought that 😅

packages/base/src/3dview/mainview.tsx
```diff
@@ -996,8 +996,12 @@ export class MainView extends React.Component<IProps, IStates> {
...
 
       const boundingBox = selectedMesh.getObjectByName(SELECTION_BOUNDING_BOX);
```
martinRenou, line 998: Last comment 😁 this will eventually never return anything now that you're not adding the box to the mesh, let's do only

```typescript
      const groupBoundingBox = this._meshGroup?.getObjectByName(
        SELECTION_BOUNDING_BOX
      );
```

arjxn-py: Thanks for making it better 🚀

martinRenou (approved):
🚀

PR #527, opened 2024-10-23
Grey icon in tabs

martinRenou:
As suggested by @SylvainCorlay 

Screenshot from 2024-10-23 11-31-50
Screenshot from 2024-10-23 11-31-21


martinRenou:
The docs build failure is unrelated, https://github.com/jupytercad/JupyterCAD/pull/528 should fix it

PR #529, opened 2024-10-23
Add Translation Controls

arjxn-py:
https://github.com/user-attachments/assets/2ce1827d-8186-4e03-99ae-3e10a8fcf80d

Shall partially supersede #151


martinRenou:
After some testing I think this is looking and working great.

Just one open question, do we want the transform control to be enabled for hidden objects? Happy to get your thoughts @arjxn-py @trungleduc @SylvainCorlay 

Allowing this allows that workflow:

https://github.com/user-attachments/assets/f2e891d4-a8a6-47d9-bbd3-2682dcc64b1f

arjxn-py:
> Just one open question, do we want the transform control to be enabled for hidden objects?

My first response to that would have been "No" but after looking at the workflow you've shared it looks like nice to have

trungleduc:
we might want to have an intermediate state for selected hidden object by playing with object's opacity.

SylvainCorlay:
> > Just one open question, do we want the transform control to be enabled for hidden objects?
>
> My first response to that would have been "No" but after looking at the workflow you've shared it looks like nice to have

Same here. I saw it and though "cool!".

martinRenou:
> we might want to have an intermediate state for selected hidden object by playing with object's opacity

I love it. 

Though transparency can be tricky, but if it's just one object at a time that should be fine.

packages/base/src/3dview/mainview.tsx
```diff
@@ -1019,7 +1021,16 @@ export class MainView extends React.Component<IProps, IStates> {
...
 
+  private _previousSelection: { [key: string]: ISelection } | null = null;
   private _updateSelected(selection: { [key: string]: ISelection }) {
+    const selectionChanged =
+      JSON.stringify(selection) !== JSON.stringify(this._previousSelection);
+
+    if (!selectionChanged) {
+      return;
+    }
+    this._previousSelection = { ...selection };
```
trungleduc, line 1032: who do we need this logic? If we really need to filter out this case, maybe it's more appropriate to do it in the `_onClick` method and use `_selectedMeshes` property.

packages/base/src/3dview/mainview.tsx
```diff
@@ -1097,10 +1115,71 @@ export class MainView extends React.Component<IProps, IStates> {
...
+              positionArray[1],
+              positionArray[2]
+            );
+
+            this._transformControls.position.copy(positionVector);
+          }
+          this._transformControls.setMode('translate');
+          this._transformControls.visible = true;
+          this._transformControls.enabled = true;
+        }
```
martinRenou, line 1141: I think we should not do any of this if there is more than one object selected. As we won't be able to translate multiple objects at once.

So this code should not be in the for loop `for (const selectionName in selection) {` but outside of it, wrapped in a `if (selection.length === 1)` check

arjxn-py: Done in [`0f65631`](https://github.com/jupytercad/JupyterCAD/pull/529/commits/0f656311ae9b5154dfe7da62177e66f5fe453829) thanks @martinRenou 🚀

packages/base/src/3dview/mainview.tsx
martinRenou: It is probably good to have a watchdog here to prevent going further updating the selection when it has not changed.

However, it may be good to track what it calling this so many times. I wonder if it could be a tiny bit related to #560

arjxn-py: > I wonder if it could be a tiny bit related to #560

Yes it can most likely be the case looking at the frequency selection is changing in #560, though can't be said with full certainty since we don't have a definite way to reproduce it yet

> it may be good to track what is calling this so many times.

Yes, i'll be on it

trungleduc: Thanks

arjxn-py: It is happening here:

https://github.com/jupytercad/JupyterCAD/blob/4f7c1fc5752624155257d9a74d8ebb5fa7d721fe/packages/base/src/3dview/mainview.tsx#L1179-L1181

arjxn-py:
> we might want to have an intermediate state for selected hidden object by playing with object's opacity.

Just this remaining, if i'm not forgetting anything (for translation)

PR #538, opened 2024-10-25
Fix clip plane controls when switching camera projection

martinRenou:
@arjxn-py Note that you will need to do the same on your transform control


trungleduc (approved):

arjxn-py:
> @arjxn-py Note that you will need to do the same on your transform control

thanks 🚀, will make sure to include this.

Issue #477, opened 2024-10-15
Object properties inspector should probably not include the filename.


SylvainCorlay:
To be consistent with the notebook property inspector.

SylvainCorlay:
Thinking about it more:

Maybe it should display the name of the selected object in the tree instead of the document name.

martinRenou:
> Maybe it should display the name of the selected object in the tree instead of the document name.

Yep!

PR #540, opened 2024-10-25
Object properties panel: Use selected object name as the title

martinRenou:
Fix https://github.com/jupytercad/JupyterCAD/issues/477
Closes https://github.com/jupytercad/JupyterCAD/issues/543

https://github.com/user-attachments/assets/b85dedbe-c056-478e-8ea1-4181b500f6f8

https://github.com/user-attachments/assets/b177e44a-291e-44d2-b0fa-feb343afd1e6


martinRenou:
Completely empty, as in removing the "object properties" title?

Screenshot from 2024-10-25 16-48-55

SylvainCorlay:
Yes, or even like in the standard panel.

Screenshot 2024-10-25 at 16 56 40

SylvainCorlay:
The current behavior is that you can still edit the properties of the last thing that was selected, which may be misleading.

trungleduc:
Clicking on other documents removes the form but still keeps the object name.

cad22

SylvainCorlay:
I think it is a clear improvement over the current state. I would be super keen on integrating it as it stands.

martinRenou:
I marked the PR as ready to review again. Although trung's comment may be an actual bug of my implementation

trungleduc:
Bot please update snapshots!

trungleduc (approved):

Issue #487, opened 2024-10-16
On changing the selection when rotating/panning with the mouse


SylvainCorlay:
I suggest we distinguish between a single 'click' for selecting an object and a 'click-and-hold' for actions such as rotating the view.

- When rotating the view, if you start and end the click-and-hold action on the same object, it should not select a new object.
- To deselect all objects, I propose that clicking on the background should suffice.

PR #542, opened 2024-10-25
Improve selection logic

martinRenou:
Fix https://github.com/jupytercad/JupyterCAD/issues/487

This was tricky to get right with the flags, but that seems to work nicely!

https://github.com/user-attachments/assets/3490d733-1e0f-41a6-9e21-f03c3e2f21fc


SylvainCorlay:
Awesome.

PR #493, opened 2024-10-17
refLength was not always updating

martinRenou:
The refLength update wasn't always done. This had the impact of eventually breaking the clipplane effect which depends on it.

PR #559, opened 2024-11-05
Do not update the camera position automatically for single object scene

martinRenou:
This was introduced in https://github.com/jupytercad/JupyterCAD/pull/493 as I thought it would be nice to move the camera to a proper position when e.g. scaling up a single object.
It turns out this can be really annoying especially when we merge [the transform control PR](#151) as it makes the camera jump upon translation (see screencast)

https://github.com/user-attachments/assets/013132e8-3c0c-4f89-86a4-af00cb31de74


arjxn-py (approved):
Thanks @martinRenou, Sounds Good

arjxn-py:
Please update snapshots

PR #564, opened 2024-11-06
Fix selection through clip plane

### LINKED ISSUE #561: Disable selecting backfaces (faces through the clip plane)
It seems we're able to select the back faces (the back of the mesh triangles) through the clip plane. It is very likely that this was always the case, but it is now visible because of the introduction of the selected object name in the object properties panel:

https://github.com/user-attachments/assets/e44deac3-fab0-4f8e-9f55-790c4fc3ee97

We should disable this in the picking logic of the main view.

martinRenou:
Fix #561 

https://github.com/user-attachments/assets/f0b2689a-202f-40c6-b291-67966e6c8025

### DIFF
--- a/packages/base/src/3dview/mainview.tsx
+++ b/packages/base/src/3dview/mainview.tsx
@@ -614,11 +614,17 @@ export class MainView extends React.Component<IProps, IStates> {
         ) {
           continue;
         }
-        const intersectMesh = intersect.object.name.includes('-front')
-          ? intersect.object.parent.getObjectByName(
-              intersect.object.name.replace('-front', '')
-            )
-          : intersect.object;
+        let intersectMesh = intersect.object;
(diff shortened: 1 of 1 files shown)


packages/base/src/3dview/mainview.tsx
```diff
@@ -614,11 +614,17 @@ export class MainView extends React.Component<IProps, IStates> {
         ) {
           continue;
         }
-        const intersectMesh = intersect.object.name.includes('-front')
-          ? intersect.object.parent.getObjectByName(
-              intersect.object.name.replace('-front', '')
-            )
-          : intersect.object;
+        let intersectMesh = intersect.object;
+        if (intersect.object.name.includes('-front')) {
```
martinRenou, line 618: Relying on the name is a bit of a poor choice. We should have UUIDs for object names on the long term to prevent any issue.

PR #565, opened 2024-11-06
Allow transforming objects when clipping is enabled

martinRenou:
This is done by having two transform controls, one specific to the clip plane, one specific to moving objects in the scene.

https://github.com/user-attachments/assets/15095e5c-cb27-405e-8a1b-e205cc6140cf


arjxn-py (approved):
Thanks @martinRenou, this looks so good.
Love it 🚀

PR #566, opened 2024-11-06
Fix initial selected edges color

martinRenou:
Tiny bug that was not hitting too much, but the initial color for selected edges was not correct.


arjxn-py (approved):
Thanks :)
Looks good.

PR #571, opened 2024-11-07
Bug fix: Prevent tree from wrongly updating the selection

### LINKED ISSUE #560: Selection state seems to jump between tree and mainview
I've seen this happen many times in my last couple of times playing with jupytercad:

https://github.com/user-attachments/assets/5e50cfca-8e0f-45e1-ad28-9ccfa03c6daf

martinRenou:
Fix #560 

Fix:
- The tree was setting the wrong selection
- The tree was updating the selection and re-rendering itself **for each mouse move** in the 3D view

This PR should prevent jumping selection and improve performance, hopefully.

### DIFF
--- a/packages/base/src/commands.ts
+++ b/packages/base/src/commands.ts
@@ -240,13 +240,11 @@ export async function executeOperator(
   }
   sharedModel.transact(() => {
     transaction(sharedModel);
-  });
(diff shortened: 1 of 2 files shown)


trungleduc:
Thanks!

Issue #572, opened 2024-11-07
Perform a dry run for any update of the shared model


martinRenou:
https://github.com/jupytercad/JupyterCAD/pull/349 introduced a "dry run" logic for operators, allowing to catch errors early and reporting them to users when an operator was not possible to perform.

This was also done in the object properties panel allowing to test changes in the shared model prior to applying them.

We should apply a "dry run" for each transaction on the shared model. **This is currently not done when moving shapes with the transform controls**.

This can lead to: 

https://gitter.ems.host/_matrix/media/v3/download/gitter.im/a0b627354abaea0bc71c2b26763d2052b2ca993b1854573277599498240?allow_redirect=true

PR #575, opened 2024-11-08
Perform a dry run for transforms

martinRenou:
Fix https://github.com/jupytercad/JupyterCAD/issues/572

This prevents a critical crash when the translation transform cannot be applied, by not applying the transform.

https://github.com/user-attachments/assets/bd61e19c-a969-4ba1-af48-cbe85489866d

### DIFF
--- a/packages/base/src/3dview/mainview.tsx
+++ b/packages/base/src/3dview/mainview.tsx
@@ -429,16 +429,11 @@ export class MainView extends React.Component<IProps, IStates> {
             positionArray[2] + updatedPosition.z
           ];
 
-          this._model.sharedModel.updateObjectByName(objectName, {
-            data: {
-              key: 'parameters',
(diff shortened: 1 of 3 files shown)


martinRenou:
@trungleduc on the long term, I guess we will want to move the dry run logic (hence the worker) higher in the stack instead of in the main view model.

That way the object properties panel does not depend on a main view model being around?

trungleduc:
> @trungleduc on the long term, I guess we will want to move the dry run logic (hence the worker) higher in the stack instead of in the main view model.
> 
> That way the object properties panel does not depend on a main view model being around?

Indeed, the main view is doing too much thing

trungleduc (approved):

arjxn-py (approved):
Thanks @martinRenou 🚀

Issue #573, opened 2024-11-07
Allow configurating the viewport


trungleduc:
Always displaying the transform control makes it easy to accidentally move the objects while rotating the camera. Other CAD software do not activate the transform control by default. I think we should follow this behavior.

Since we had already the dialog to show and hide the axes helper, we can change this dialog to a viewport config dialog with the following items:
- Toggle axes helper
- Toggle object highlight mode:  using a bounding box or using highlight color
- Highlight or bounding box color selector
- Viewport background color, it would be great if we could support gradient color.

We could reuse the setting system of JupyterLab to persist the viewport config.

arjxn-py:
Makes Sense! Will be happy to do it :)

trungleduc:
I updated the issue about the viewport config. The transform control merits its own toggle button on the toolbar

martinRenou:
> - Toggle object highlight mode: using a bounding box or using highlight color
> - Highlight or bounding box color selector
> - Viewport background color, it would be great if we could support gradient color

I'm thinking these should be using normal JupyterLab settings.

There may be a way to add a shortcut in the toolbar to go to the settings page for JupyterCAD?

> - Toggle axes helper

I feel like this one should instead be stored on a per-file basis in StateDB like the other view states listed in https://github.com/jupytercad/JupyterCAD/issues/601

trungleduc:
> I'm thinking these should be using normal JupyterLab settings.

> There may be a way to add a shortcut in the toolbar to go to the settings page for JupyterCAD?

If we put it into JupyterLab settings, having the shortcut is a must. I know lots of users do not know about the setting page of JupyterLab.

> I feel like this one should instead be stored on a per-file basis in StateDB like the other view states listed in https://github.com/jupytercad/JupyterCAD/issues/601

just checked FreeCAD, it's at application level. I would follow them for these configs

PR #578, opened 2024-11-11
Enable toggling transform controls

arjxn-py:
https://github.com/user-attachments/assets/97e25dfd-6135-42d9-bd0a-2f5b8610f157

Icon is a draft
In the direction of #573


trungleduc:
Thanks!

packages/base/src/3dview/mainview.tsx
```diff
@@ -1455,6 +1457,20 @@ export class MainView extends React.Component<IProps, IStates> {
...
     }
+
+    if (change.key === 'transform') {
+      const transformEnabled = change.newValue as boolean | undefined;
+
+      if (transformEnabled !== undefined) {
+        this.setState(
+          old => ({ ...old, transform: transformEnabled }),
+          () => {
+            this._updateSelected({});
```
martinRenou, line 1468: Is this right? Shouldn't we just set the transformControls object's `visible` attribute here instead? Or the `enabled` property

martinRenou (changes requested):
Just have a comment/question

packages/base/src/3dview/mainview.tsx
arjxn-py: yes, i tried that but just setting those was preventing transformControls from disabling or enabling right on clicking the toggle button, which is why I also had to trigger this part of code: 

https://github.com/jupytercad/JupyterCAD/blob/3340d1d62c30b4c863dac74563bbaec19f26db5e/packages/base/src/3dview/mainview.tsx#L1181-L1196

martinRenou: Hmm, though calling `_updateSelected({})` means unselecting anything that was previously selected.

I'm a bit confused you need this code to be triggered again. This code should already have been triggered upon the latest selection.

arjxn-py: > This code should already have been triggered upon the latest selection.

i am needed to reselect the object manually after toggling the transform controls from the button 

> though calling _updateSelected({}) means unselecting anything that was previously selected.

yes this is not ideal and should be done better, happy to hear thoughts

arjxn-py: > > i am needed to reselect the object manually
> 
> Do you know why you need that?

It was most probably needed because of missing attach & detach calls.
Is it better after [`f0b2c6d`](https://github.com/jupytercad/JupyterCAD/pull/578/commits/f0b2c6d28af1a0ee2bd271f68c37718a7c26b8a5)? if not what should be the right method to get the selected object?

trungleduc:
Could you append a `div` containing the helper text to switch between rotation and translation mode into the main view div element? You could use the absolute position to place it in the bottom right of the viewport.

arjxn-py:
> Could you append a `div` containing the helper text to switch between rotation and translation mode into the main view div element? You could use the absolute position to place it in the bottom right of the viewport. Like this one

Yes, that looks great. Thank You 🚀 i'll be on it

packages/base/src/3dview/mainview.tsx
martinRenou: That does not look correct to me `this._pick()` will run the raycast picking, so it's entirely dependent on your current mouse position. This means your code does not work (won't attach to the correct object) if the mouse is mal-positioned.

I'm still not sure why you need to attach/detach here?

[It should already have been attached here on the latest selection](https://github.com/jupytercad/JupyterCAD/blob/9b4160bf274697db9783892dd56010b315399f0a/packages/base/src/3dview/mainview.tsx#L1185). So AFAIU your code here should only do `enabled = false/true` and `visible = false/true`.

trungleduc:
This element can also be used for the clipping plane

packages/base/src/3dview/mainview.tsx
arjxn-py: Ahh, now it works without attach/detach calls, i must not have been doing it 100% correctly before

martinRenou: Thanks for checking this out!

packages/base/src/3dview/mainview.tsx
```diff
@@ -1429,9 +1435,15 @@ export class MainView extends React.Component<IProps, IStates> {
...
       if (change.type !== 'remove' && clipSettings) {
-        this._clipSettings = clipSettings;
+        this.setState(
+          oldState => ({ ...oldState, clipEnabled: clipSettings.enabled }),
+          () => {
+            this._clipSettings = clipSettings;
+            this._updateClipping();
 
-        this._updateClipping();
+            this._renderer.render(this._scene, this._camera);
```
martinRenou, line 1444: Do we really need to call `render`?

packages/base/src/3dview/mainview.tsx
```diff
@@ -1455,6 +1467,27 @@ export class MainView extends React.Component<IProps, IStates> {
...
+          () => {
+            if (transformEnabled) {
+              this._transformControls.visible = true;
+              this._transformControls.enabled = true;
+            } else {
+              this._transformControls.visible = false;
+              this._transformControls.enabled = false;
+            }
+
+            this._renderer.render(this._scene, this._camera);
```
martinRenou, line 1486: Same question here

martinRenou (changes requested):
Latest question concerning the extra call to `renderer.render`, I believe the rendering loop should already render the scene on the next animation frame request, so it should not be needed.

Otherwise it looks great thanks!

packages/base/src/3dview/mainview.tsx
arjxn-py: It works without it, Thanks :)

arjxn-py:
> It's done but strangely it doesn't show up for clip plane on first toggle, after first toggle it works perfectly though

Just one concern i've remaining.

packages/base/src/3dview/mainview.tsx
```diff
@@ -1449,7 +1459,25 @@ export class MainView extends React.Component<IProps, IStates> {
...
+        this.setState(
+          old => ({ ...old, transform: transformEnabled }),
+          () => {
+            if (transformEnabled) {
+              this._transformControls.visible = true;
+              this._transformControls.enabled = true;
+            } else {
+              this._transformControls.visible = false;
+              this._transformControls.enabled = false;
             }
```
martinRenou, line 1481: Nitpicking a bit 🧹 

```suggestion
            this._transformControls.visible = transformEnabled;
            this._transformControls.enabled = transformEnabled;
```

arjxn-py: I should've saw it 🫠

martinRenou (approved):
Thanks!

PR #594, opened 2024-11-28
Disable transform control for readonly files

### LINKED ISSUE #592: Should the transform controls be disabled for STEP/STL files like other buttons changing the model
At the moment, we can still show the transform control for STL and STEP files.

martinRenou:
Fix #592


arjxn-py (approved):
Thanks

SylvainCorlay:
If we make these things command-based, we should eventually make the comment available or unavailable depending on the context.

PR #600, opened 2024-11-28
Fix missing placement for STL & STEP Files

arjxn-py:


martinRenou (approved):
Thanks!

martinRenou:
Not waiting for more CI, it does not test this part of the code anyway

PR #574, opened 2024-11-07
Transformation Controls - Rotation

arjxn-py:

PR #603, opened 2024-11-28
Rotation Controls

arjxn-py:
Shall supersede #574


packages/base/src/3dview/mainview.tsx
```diff
@@ -724,18 +740,21 @@ export class MainView extends React.Component<IProps, IStates> {
...
-            this._clipPlaneTransformControls.setMode('rotate');
-          }
-          break;
+      if (event.key === 'r' && this._clipSettings.enabled) {
+        event.preventDefault();
+        event.stopPropagation();
+        toggleMode(this._clipPlaneTransformControls);
+      }
+
+      if (event.key === 't' && this._transformControls.enabled) {
```
martinRenou, line 754:
```suggestion
      if (event.key === 'r' && this._transformControls.enabled) {
```

martinRenou (commented):
Let's make it 'r' too

trungleduc:
I get this error


https://github.com/user-attachments/assets/c7c405ac-fea0-4b3d-9703-ce1fdd2df1aa

martinRenou:
> I get this error

Also had that one once, but not after. We should check the value set in the shared model (possible NaN?)

packages/base/src/3dview/mainview.tsx
```diff
@@ -423,6 +423,20 @@ export class MainView extends React.Component<IProps, IStates> {
 
         const updatedPosition = new THREE.Vector3();
         updatedObject.getWorldPosition(updatedPosition);
+        const updatedQuaternion = new THREE.Quaternion();
+        updatedObject.getWorldQuaternion(updatedQuaternion);
+
+        const s = Math.sqrt(1 - updatedQuaternion.w * updatedQuaternion.w);
+        const updatedRotation = [
+          [
+            parseFloat((updatedQuaternion.x / s).toFixed(10)),
```
martinRenou, line 432: I suspect the error comes from `parseFloat` which returns may `NaN`. Do we really need `parseFloat` here? Why not just `updatedQuaternion.x / s`?

martinRenou: Probably a division by `0` when s is `0`, there should be a check for this

arjxn-py: > Do we really need `parseFloat` here? Why not just `updatedQuaternion.x / s`?

Yes, it is also sufficient.

> Probably a division by 0 when s is 0, there should be a check for this

Added a fallback condition, thanks :)

arjxn-py:
> I get this error

the NaN shall be fixed by [`9387bcf`](https://github.com/jupytercad/JupyterCAD/pull/603/commits/9387bcf70b5997f5f4ecacd13b7996a84ecee6cd)

trungleduc:
Thanks! it works

trungleduc (approved):

martinRenou:
Yay!

martinRenou (approved):

PR #610, opened 2024-11-29
Transform controls: undo positioning upon failure

### LINKED ISSUE #588: After dry-run failure, we should bring back the MeshGroup's original position
See this screencast of a cut of two shapes followed by a chamfer, moving the original box out will result in not being able to compute the chamfer, but we don't recover the initial box's position

https://github.com/user-attachments/assets/c0f91b6d-27c7-4745-9f5f-694285b964c8

martinRenou:
Fix #588 

https://github.com/user-attachments/assets/17f7cdf7-7b09-484b-83da-a8157dae07b8

### DIFF
--- a/packages/base/src/3dview/helpers.ts
+++ b/packages/base/src/3dview/helpers.ts
@@ -71,6 +71,10 @@ export interface IMeshGroupMetadata {
   [key: string]: any;
 }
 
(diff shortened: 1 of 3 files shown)


arjxn-py (approved):
Thanks @martinRenou 🚀

PR #619, opened 2024-12-02
Enable renaming object name from Object Tree

### LINKED ISSUE #609: Allow renaming objects from the tree view?
This may need https://github.com/jupytercad/JupyterCAD/issues/200

arjxn-py:
Shall close #609 

https://github.com/user-attachments/assets/4a969e15-c787-44e5-9b2f-bf70272ff9dd

### DIFF
--- a/packages/base/src/panelview/objecttree.tsx
+++ b/packages/base/src/panelview/objecttree.tsx
@@ -89,6 +89,8 @@ interface IStates {
   selectedNodes: string[];
   clientId: number | null; // ID of the yjs client
   id: string; // ID of the component, it is used to identify which component
+  editingNode: TreeNodeId | null;
+  editingNodeValue: string | null;
   //is the source of awareness updates.
   openNodes: (number | string)[];
 }
@@ -154,7 +156,9 @@ class ObjectTreeReact extends React.Component<IProps, IStates> {
       selectedNodes: [],
       clientId: null,
       id: uuid(),
-      openNodes: []
+      openNodes: [],
+      editingNode: null,
+      editingNodeValue: ''
     };
     this.props.cpModel.jcadModel?.sharedObjectsChanged.connect(
       this._sharedJcadModelChanged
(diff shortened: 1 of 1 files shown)


arjxn-py:
There's a small issue

packages/base/src/panelview/objecttree.tsx
```diff
@@ -464,8 +468,61 @@ class ObjectTreeReact extends React.Component<IProps, IStates> {
...
+                            const updatedName = this.state.editingNodeValue;
+                            const objectName = opts.node.id;
+
+                            if (this.props.cpModel.mainViewModel) {
+                              const obj =
+                                this.props.cpModel.jcadModel?.sharedModel.getObjectByName(
+                                  objectName as string
+                                );
+                              if (obj) {
+                                this.props.cpModel.sharedModel?.updateObjectByName(
```
martinRenou, line 497: We not only need to update the specific object, but we also need to go through all objects and see if they depend on it. All objects have a `dependency` parameter which is a list of object it depends on, by name.

I believe it's easier to tackle this renaming task once https://github.com/jupytercad/JupyterCAD/issues/200 is done. We wouldn't need to do that work I'm describing above because the list of dependencies would be a list of UUIDs.

I would suggest we put this renaming PR on hold while https://github.com/jupytercad/JupyterCAD/issues/200 is not tackled.

martinRenou (changes requested):
Thanks a lot for starting this! 🚀 

I we should tackle https://github.com/jupytercad/JupyterCAD/issues/200 first.

arjxn-py:
> And you need to propagate the change to all downstream operators.

Yeah, I didn't thought about it, i agree with @martinRenou to tackle #200 first

martinRenou:
@arjxn-py do you mind if we revert the branch removal and reopen this PR as draft? Just to keep it visible. The diff will be useful in time.

arjxn-py:
Oh oops, I deleted some branches from my fork in bulk and this one must have been deleted by mistake.
I'm going to revert the deletion and keeping the PR open as draft

PR #623, opened 2024-12-02
Introducing file versioning

martinRenou:


martinRenou:
We would need to improve error handling jupyter-collaboration's side, as it's hiding the actual error we are raising:

Screenshot From 2024-12-02 12-35-09

The raised error shows up in the server logs only.

python/jupytercad_core/jupytercad_core/jcad_ydoc.py
```diff
@@ -42,6 +44,12 @@ def set(self, value: str) -> None:
         :type value: Any
         """
         valueDict = json.loads(value)
+
+        # Assuming file version 3.0.0 if the version is not specified
+        file_version = valueDict["schemaVersion"] if "schemaVersion" in valueDict else "3.0.0"
+        if file_version != CURRENT_SCHEMA_VERSION:
+            raise ValueError(f"Cannot load file version {file_version}")
```
martinRenou, line 59: I'm raising here, because this is the place we will be reading and adapting the file content to the current schema (e.g. adapting file 3.0.0 to to 4.0.0 in the future).

Thoughts @trungleduc ?

trungleduc: A safer bet is to compare the version? We should support any version lower than the current schema version

martinRenou: > We should support any version lower than the current schema version

I guess there could be a day when we drop support for files that are too old if we have to. But yes comparing the versions sounds good for now.

python/jupytercad_core/jupytercad_core/constants.py
```diff
@@ -0,0 +1 @@
+CURRENT_SCHEMA_VERSION = "3.0.0"
```
trungleduc, line 1: ideally, we could generate files containing this value from the value in the `jcad.json` file, this will ensure a single source of trust for the schema version. But I'm fine defining them manually.

python/jupytercad_core/jupytercad_core/jcad_ydoc.py
```diff
@@ -42,6 +50,14 @@ def set(self, value: str) -> None:
         :type value: Any
         """
         valueDict = json.loads(value)
+
+        # Assuming file version 3.0.0 if the version is not specified
+        file_version = (
+            valueDict["schemaVersion"] if "schemaVersion" in valueDict else "3.0.0"
```
trungleduc:
```suggestion
            valueDict["schemaVersion"] if "schemaVersion" in valueDict else CURRENT_SCHEMA_VERSION
```

trungleduc: Actually keeping it as `3.0.0` makes more sense

python/jupytercad_core/jupytercad_core/constants.py
martinRenou: Indeed, I can access that value from 

```python
from jupytercad_lab.notebook.objects._schema.jcad import IJCadContent
IJCadContent.model_fields['schemaVersion'].default
```

If we take that route though it would be cleaner to have the generated Python classes be part of `jupytercad_core`, any objection doing this?

martinRenou: When it comes to the TypeScript generated interfaces though, it does not have the schema value information that comes from `jcad.json`. Not sure how to workaround that in that case.

trungleduc: Indeed it's cleaner.

trungleduc: there is a node function in `packages/schema/schema.json` which reads the json files, you can use it to write your constant somewhere

martinRenou: Sounds good

martinRenou: I've pushed a commit doing all this :)

packages/schema/src/doc.ts
```diff
@@ -4,6 +4,8 @@ import { ISignal, Signal } from '@lumino/signaling';
 import * as Y from 'yjs';
 
 import { IJCadObject, IJCadOptions } from './_interface/jcad';
+export { SCHEMA_VERSION } from './_interface/version';
```
trungleduc, line 7: it can be exported directly in `index.ts`?

martinRenou: Sounds good!

trungleduc (approved):

arjxn-py:
Done in https://github.com/geojupyter/jupytergis/pull/663

PR #653, opened 2024-12-13
Fix placement for non-primitive shapes coming from freecad

### LINKED ISSUE #651: Rotation transforms not working with the Gear example
The object appears to render, but appears in the original position after releasing the mouse.

arjxn-py:
Shall fix #651

### DIFF
--- a/packages/occ-worker/src/occapi/any.ts
+++ b/packages/occ-worker/src/occapi/any.ts
@@ -1,7 +1,6 @@
 import { OCC } from '@jupytercad/opencascade';
 import { IAny, IJCadContent } from '@jupytercad/schema';
 
-import { setShapePlacement } from './common';
 import { _loadObjectFile } from './loadObjectFile';
 
 export function _Any(
@@ -14,7 +13,5 @@ export function _Any(
     type: Type,
     placement: Placement
   });
-  if (result) {
-    return setShapePlacement(result, Placement);
-  }
+  return result;
 }
--- a/packages/occ-worker/src/occapi/any.ts
+++ b/packages/occ-worker/src/occapi/any.ts
@@ -9,7 +9,11 @@ export function _Any(
   content: IJCadContent
 ): OCC.TopoDS_Shape | undefined {
   const { Content, Type, Placement } = arg;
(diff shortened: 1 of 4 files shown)


packages/occ-worker/src/occapi/any.ts
```diff
@@ -9,7 +9,11 @@ export function _Any(
...
   const { Content, Type, Placement } = arg;
-  const result = _loadObjectFile({ content: Content, type: Type });
+  const result = _loadObjectFile({
+    content: Content,
+    type: Type,
+    placement: Placement
+  });
   if (result) {
     return setShapePlacement(result, Placement);
   }
```
martinRenou, line 19: We don't need this anymore since `_loadObjectFile` is already setting the placement correctly now

packages/occ-worker/src/occapi/index.ts
```diff
@@ -66,6 +66,11 @@ export const Fillet = operatorCache<IFillet>('Part::Fillet', _Fillet);
 export const ObjectFile = operatorCache<{
   content: string;
   type: IAny['Type'];
+  placement: {
```
trungleduc, line 69:
```suggestion
  placement?: {
```

trungleduc: The generic type parameter here takes the same type of the first argument of `_loadObjectFile`

arjxn-py: > The generic type parameter here takes the same type of the first argument of `_loadObjectFile`

It is of same type as type in `_loadObjectFile`, do we want to change?

packages/occ-worker/src/occapi/any.ts
arjxn-py: Done, Thanks :)

martinRenou (approved):
Thanks!

trungleduc (approved):

martinRenou:
https://github.com/user-attachments/assets/e5409344-11c9-44e2-b7c9-4e5414430cec

Confirming it works in the JupyterLite deployment!

PR #658, opened 2025-01-01
Use `buildShape` in `objToMesh`

### LINKED ISSUE #605: objToMesh should use the buildShape helper
The function objToMesh: https://github.com/jupytercad/JupyterCAD/blob/main/packages/base/src/3dview/mainview.tsx#L923
 
is not correctly implemented. It should use the `buildShape` helper function. 
The downside of not using `buildShape` is:
- wrong coloring 
- no bounding box for selection
- no support for clip plane
- etc

arjxn-py:
Shall close #605


packages/base/src/3dview/mainview.tsx
```diff
@@ -996,38 +997,65 @@ export class MainView extends React.Component<IProps, IStates> {
...
+        );
+        mainMesh.material.color = originalMeshColor;
+        mainMesh.userData.originalColor = originalMeshColor.clone();
+      }
+
+      edgesMeshes.forEach(edgeMesh => {
+        this._edgeMaterials.push(edgeMesh.material);
+        const edgeColor = new THREE.Color(objColor || DEFAULT_EDGE_COLOR);
+        edgeMesh.material.color = edgeColor;
+      });
```
martinRenou, line 1056: It seems weird to me we have to do this?

Shouldn't this be done in `buildShape`?

PR #659, opened 2025-01-02
Fuse multiple objects at once 

### LINKED ISSUE #644: Cannot fuse an entire selection at once
We can only fuse two objects at a time even though the OpenCascade routine and the schema we're using for Intersection can manage more than two input objects.

https://github.com/user-attachments/assets/6334e4cf-039d-464f-9273-0b5cf7d8eb70

arjxn-py:
Should fix #644 

https://github.com/user-attachments/assets/d5560387-0432-4570-bcc4-a6c6e80d98b6


arjxn-py:
The fuse form is now considering multiple selection but the result of fuse is still considering the first two selected shapes!

packages/base/src/commands.ts
```diff
@@ -345,13 +345,21 @@ const OPERATORS = {
...
+      const selectedShapes = Object.keys(selected).map(key => key);
+
+      // Fallback to at least two objects if selection is empty
+      const baseShapes =
+        selectedShapes.length > 0
+          ? selectedShapes
+          : [objects[0].name || '', objects[1].name || ''];
+
+      console.log('baseShapes', baseShapes);
+
```
arjxn-py, line 358:
```suggestion
```

trungleduc (approved):

trungleduc:
Thanks!

PR #677, opened 2025-02-06
Fix shared document interface

arjxn-py:
Porting https://github.com/geojupyter/jupytergis/pull/406

### DIFF
--- a/packages/schema/src/interfaces.ts
+++ b/packages/schema/src/interfaces.ts
@@ -109,6 +109,9 @@ export interface IJupyterCadDoc extends YDocument<IJupyterCadDocChange> {
   setOutput(key: string, value: IPostResult): void;
   removeOutput(key: string): void;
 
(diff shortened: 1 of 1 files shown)


martinRenou (approved):
Thanks!

martinRenou:
It looks like https://github.com/jupytercad/JupyterCAD/blob/main/packages/schema/src/doc.ts#L69 should be using `JSONObject`

arjxn-py:
> It looks like [`main`/packages/schema/src/doc.ts#L69](https://github.com/jupytercad/JupyterCAD/blob/main/packages/schema/src/doc.ts?rgh-link-date=2025-02-06T10%3A57%3A39Z#L69) should be using `JSONObject`

Thanks. Yes, i'm on it :)

martinRenou:
Mmh, you should check `setSource` too. Not sure why TypeScript doesn't complain.

PR #678, opened 2025-02-06
Create a new file from the Python API

arjxn-py:
Ports https://github.com/geojupyter/jupytergis/pull/402


martinRenou:
As discussed in private, we should hold before merging this. The choice to create the file from Python is not correct since only the front-end will know the correct drive (in the sense of JupyterLab drive) where to create the file.

martinRenou:
Creation of empty files is handled in JS by https://github.com/jupytercad/JupyterCAD/pull/680

PR #679, opened 2025-02-10
Widget with toolbar and sidepanel in cell output

arjxn-py:
Porting https://github.com/geojupyter/jupytergis/pull/419

widget


packages/base/src/panelview/leftpanel.tsx
```diff
@@ -34,16 +33,16 @@ export class LeftPanelWidget extends SidePanel {
     this.addWidget(properties);
 
     this._handleFileChange = () => {
-      header.title.label = this._currentContext?.localPath || '-';
+      header.title.label = this._currentContext?.filePath || '-';
```
martinRenou, line 36: I'd rename this property, since it's the model now

```suggestion
      header.title.label = this._currentModel?.filePath || '-';
```

packages/base/src/panelview/leftpanel.tsx
```diff
@@ -34,16 +33,16 @@ export class LeftPanelWidget extends SidePanel {
...
     };
 
     options.tracker.currentChanged.connect((_, changed) => {
       if (changed) {
         if (this._currentContext) {
           this._currentContext.pathChanged.disconnect(this._handleFileChange);
         }
```
martinRenou, line 43: Should we remove these? Since the `model.filePath` is not something that changes?

```suggestion
```

packages/base/src/panelview/leftpanel.tsx
```diff
@@ -34,16 +33,16 @@ export class LeftPanelWidget extends SidePanel {
...
-        this._currentContext = changed.context;
-        header.title.label = changed.context.localPath;
+        this._currentContext = changed.model;
+        header.title.label = changed.model.filePath;
         this._currentContext.pathChanged.connect(this._handleFileChange);
```
martinRenou, line 46: Same here?

```suggestion
```

packages/base/src/panelview/rightpanel.tsx
```diff
@@ -45,29 +45,29 @@ export class RightPanelWidget extends SidePanel {
...
     this._handleFileChange = () => {
-      header.title.label = this._currentContext?.localPath || '-';
+      header.title.label = this._currentContext?.filePath || '-';
     };
 
     options.tracker.currentChanged.connect(async (_, changed) => {
       if (changed) {
         if (this._currentContext) {
           this._currentContext.pathChanged.disconnect(this._handleFileChange);
         }
```
martinRenou, line 55: Same comments for the right panel

```suggestion
```

packages/base/src/commands.ts
```diff
@@ -1177,11 +1173,11 @@ export function addCommands(
       }
 
       const dialog = new FormDialog({
-        context: current.context,
+        model: current.model,
         title: EXPORT_FORM.title,
         schema: EXPORT_FORM.schema,
-        sourceData: EXPORT_FORM.default(tracker.currentWidget?.context),
```
martinRenou, line 1183: The typings of callbacks in `EXPORT_FORM` will need to be changed too

packages/base/src/panelview/leftpanel.tsx
trungleduc: it does change when you rename the jcad file

martinRenou: Indeed. Since the model has a `filePath` property, how about we add a `pathChanged` getter on the model which returns the context `pathChanged` signal?

trungleduc:
Thanks @arjxn-py, the failed UI tests are not related?

martinRenou:
Seems related, the toolbar now shows up in the cell outputs, but it does not show up anymore in the normal case

packages/base/src/widget.ts
```diff
@@ -19,17 +21,27 @@ import {
...
+  implements IJupyterCadDocumentWidget
 {
   constructor(
     options: DocumentWidget.IOptions<JupyterCadPanel, IJupyterCadModel>
   ) {
     super(options);
   }
 
+  get model(): IJupyterCadModel {
+    return this.model;
```
martinRenou, line 42: That's an infinite loop 🏃🏽‍♂️

arjxn-py: Using vs-code replace all turned out to be a little headache this time 🤧

python/jupytercad_core/src/factory.ts
```diff
@@ -50,14 +50,15 @@ export class JupyterCadWidgetFactory extends ABCWidgetFactory<
...
-  ): JupyterCadWidget {
+  ): JupyterCadDocumentWidget {
     if (this._backendCheck) {
       const checked = this._backendCheck();
       if (!checked) {
         throw new Error('Requested backend is not installed');
       }
     }
     const { model } = context;
+    model.filePath = context.localPath;
```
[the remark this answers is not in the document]
arjxn-py: image
Listening to this disappears the file name from panels like before.

arjxn-py:
Bot please update snapshots.

python/jupytercad_core/src/factory.ts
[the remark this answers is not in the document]
martinRenou: I cannot reproduce what you're seeing Arjun. With the change above, the file rename works again:


https://github.com/user-attachments/assets/aff2b043-f309-4906-b022-c83508eaf4a2

martinRenou (approved):
Looks great 🚀 !!

PR #680, opened 2025-02-11
Enable Python API in the JupyterLite deployment

arjxn-py:
Porting https://github.com/geojupyter/jupytergis/pull/412


yarn.lock
```diff
@@ -13048,6 +13048,28 @@ __metadata:
   languageName: node
   linkType: hard
 
+"yjs-widgets@npm:^0.4":
```
martinRenou, line 13051: Mmh it looks like there are two yjs-widgets in the `yarn.lock` now. `0.4.0` and `0.3.8`.

martinRenou: It's usually not a good idea to modify the `yarn.lock` manually, what I usually do is read the `yarn.lock` to understand who is pulling the old yjs-widgets version.

And it seems like it's jupytercad-app that pulls it: https://github.com/jupytercad/JupyterCAD/blob/1474ba8d33132a0d8fedda961ef3808cc4a0768e/yarn.lock#L873

So you need to modify the `package.json` of jupytercad-app with the new pinning, then re-run `jlpm install` to update the `yarn.lock`

arjxn-py: Done.

martinRenou (approved):
It works!!!

Screenshot From 2025-02-12 10-10-00

arjxn-py:
> It works!!!

Thanks a lot @martinRenou!! Wouldn't have been possible alone :)

PR #681, opened 2025-02-12
Add `jupytercad_lite` metapackage

arjxn-py:
Porting https://github.com/geojupyter/jupytergis/pull/451

### DIFF
--- a/scripts/dev-install.py
+++ b/scripts/dev-install.py
@@ -35,10 +35,12 @@ def install_dev(no_occ_build=False):
             execute("jupyter server extension enable jupytercad_core")
 
         if py_package != "jupytercad_app":
-            execute(f"pip install -e {python_package_prefix}/jupytercad")
             execute(
                 f"jupyter labextension develop {python_package_prefix}/{py_package} --overwrite"
             )
+
+    if py_package != "jupytercad_app":
+        execute(f"pip install -e {python_package_prefix}/jupytercad")
 
 
 if __name__ == "__main__":
--- a/scripts/dev-install.py
+++ b/scripts/dev-install.py
@@ -35,9 +35,7 @@ def install_dev(no_occ_build=False):
             execute("jupyter server extension enable jupytercad_core")
 
         if py_package != "jupytercad_app":
-            execute(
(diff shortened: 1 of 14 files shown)


scripts/dev-install.py
```diff
@@ -35,9 +35,7 @@ def install_dev(no_occ_build=False):
             execute("jupyter server extension enable jupytercad_core")
 
         if py_package != "jupytercad_app":
-            execute(
-                f"jupyter labextension develop {python_package_prefix}/{py_package} --overwrite"
-            )
+            execute(f"pip install -e {python_package_prefix}/jupytercad")
```
martinRenou, line 38: I'm confused about this change? We should continue installing the labextension in the dev install script

arjxn-py: Ahh, i just copied this from jupytergis PR, should i revert?

martinRenou: I should review my own PRs 😆 I should not have removed it in the original PR

martinRenou: so, the `jupyter labextension develop` command should stay, and the `execute(f"pip install -e {python_package_prefix}/jupytercad")` should stay too

arjxn-py: Done :)

scripts/dev-install.py
```diff
@@ -35,6 +35,7 @@ def install_dev(no_occ_build=False):
             execute("jupyter server extension enable jupytercad_core")
 
         if py_package != "jupytercad_app":
+            execute(f"pip install -e {python_package_prefix}/jupytercad")
```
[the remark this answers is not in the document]
arjxn-py: Done, outside the loop. Thank you for pointing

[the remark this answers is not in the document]
martinRenou: You shouldn't keep the condition `if py_package != "jupytercad_app":` 😁 since you're not in the loop anymore and that condition doesn't quite make sense

martinRenou (approved):
Thanks!

martinRenou (approved):
Thanks!

PR #692, opened 2025-02-17
Add setSource in stl & step plugin to fix lite deployment

arjxn-py:


trungleduc:
it works?

python/jupytercad_core/src/stepplugin/model.ts
```diff
@@ -56,6 +57,41 @@ export class JupyterCadStepDoc extends JupyterCadDoc {
...
+      Object.entries(metadata).forEach(([key, val]) =>
+        this._metadata.set(key, val as string)
+      );
+
+      const outputs = value['outputs'] ?? {};
+      Object.entries(outputs).forEach(([key, val]) =>
+        this._outputs.set(key, val as IPostResult)
+      );
+    });
+  }
```
martinRenou, line 93: I believe this is not fixing the issue? You're still assuming the source is a JSON object here, which does not work. Shouldn't we do something along the way of:

```suggestion
  setSource(value: string): void {
    this._source.insert(0, value);
  }
```

cc. @trungleduc

martinRenou (changes requested):

arjxn-py:
> does it work when you try it locally?

yes 
image

trungleduc:
> > does it work when you try it locally?
> 
> yes image

no, i mean testing it in the lite version

python/jupytercad_core/src/stepplugin/model.ts
trungleduc: 👍

arjxn-py:
> no, i mean testing it in the lite version

I couldn't test lite deployment locally yet but was waiting for the appsharing link to show up which is now loading

trungleduc:
> > no, i mean testing it in the lite version
> 
> I couldn't test lite deployment locally yet but was waiting for the appsharing link to show up which is now loading

You can exactly follow the steps in the build Jupyterlite CI action to generate JupyterLite deployment locally.

arjxn-py:
image
This is fixed, thank you both

trungleduc (approved):

martinRenou (approved):
Thanks!

PR #696, opened 2025-02-19
Separate Bot for lite snapshot update

arjxn-py:


.github/workflows/update_lite_galata_references.yaml
```diff
@@ -0,0 +1,111 @@
...
+    shell: bash -l {0}
+
+jobs:
+  update-snapshots:
+    if: >
+      (
+        github.event.comment.author_association == 'OWNER' ||
+        github.event.comment.author_association == 'COLLABORATOR' ||
+        github.event.comment.author_association == 'MEMBER'
+      ) && github.event.issue.pull_request && contains(github.event.comment.body, 'please update litesnaps')
```
martinRenou, line 22: Let's keep the same triggering rule

```suggestion
      ) && github.event.issue.pull_request && contains(github.event.comment.body, 'please update snapshots')
```

.github/workflows/update_lite_galata_references.yaml
```diff
@@ -0,0 +1,111 @@
...
+        run: |
+          jlpm install
+          yarn playwright install
+
+      - uses: jupyterlab/maintainer-tools/.github/actions/update-snapshots@main
+        with:
+          npm_client: jlpm
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          start_server_script: 'null'
+          test_folder: ui-tests
```
martinRenou, line 111: I believe you're missing an `update_script` parameter here https://github.com/jupyterlab/maintainer-tools/blob/main/.github/actions/update-snapshots/action.yml#L58

We need a script that updates the lite snapshots specifically

.github/workflows/update_lite_galata_references.yaml
```diff
@@ -0,0 +1,111 @@
...
+      #       jupyterlab=4
+      #       pythonocc-core=7.7.0
+
+      # - name: Install the extension
+      #   shell: bash -l {0}
+      #   run: |
+      #     whereis python
+      #     cp ./jupytercad_core/dist/jupytercad*.whl ./jupytercad_lab/dist/jupytercad*.whl ./jupytercad_app/dist/jupytercad*.whl ./jupytercad/dist/jupytercad*.whl .
+      #     python -m pip install jupytercad*.whl
+
```
martinRenou, line 96: 🧹 

```suggestion
```

martinRenou (changes requested):

.github/workflows/update_lite_galata_references.yaml
arjxn-py: I wonder if it'd be a conflict

.github/workflows/update_lite_galata_references.yaml
arjxn-py: Does [`8e0ffaf`](https://github.com/jupytercad/JupyterCAD/pull/696/commits/8e0ffaf9b23cc2bd5e6929b4222244a7a9785832) does it correctly?

martinRenou (approved):
Thanks! Let's merge and try

PR #698, opened 2025-02-20
Install jupyterlab to enable jlpm on update lite snaphots bot

arjxn-py:


arjxn-py:
Merging this to test

.github/workflows/update_lite_galata_references.yaml
```diff
@@ -84,6 +84,7 @@ jobs:
         env:
           PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: 1
         run: |
+          pip install jupyterlab
```
martinRenou, line 87: Do you really have pip available at this point ? You may need to create a micromamba env like the other job is doing

arjxn-py: Although it does have installed jupyterlab with pip but it now fails at the next step - https://github.com/jupytercad/JupyterCAD/actions/runs/13434015494/job/37532083130?pr=689#step:8:1

martinRenou: And we should remove this line

arjxn-py: Yes it did install jupyterlab with pip but it failed at next step with - `Usage Error: Couldn't find a script named "test:updatelite".` - https://github.com/jupytercad/JupyterCAD/actions/runs/13434015494/job/37532083130?pr=689#step:9:1203

martinRenou: You don't seem to have a test:updatelite script yet in your PR?

PR #713, opened 2025-04-18
Use svg icons only for the toolbar icons

HaudinFlorence:
Replace the pencil, the camera and the terminal console by svg icon with the `jp-icon3 style`.
Screenshot From 2025-04-18 18-57-52
Screenshot From 2025-04-18 18-58-13


SylvainCorlay (approved):

trungleduc (changes requested):
> src/commands.ts(41,3): error TS6133: 'terminalToolbarIcon' is declared but its value is never read.

HaudinFlorence:
> > src/commands.ts(41,3): error TS6133: 'terminalToolbarIcon' is declared but its value is never read.

Thanks for the comment @trungleduc. Unused import of terminalToolbarIcon has been removed.

trungleduc (approved):

PR #717, opened 2025-04-22
Replace the formDialog by a toggle button for CameraSettings

HaudinFlorence:
Modify updateCameraSettings command to use a toogle button instead of a formDialog.

Replace the camera icon by a video camera one.

### DIFF
--- a/packages/base/src/commands.ts
+++ b/packages/base/src/commands.ts
@@ -37,7 +37,7 @@ import {
   wireframeIcon,
   transformIcon,
   pencilSolidIcon,
-  cameraSolidIcon
+  videoSolidIcon
 } from './tools';
 import keybindings from './keybindings.json';
 import { DEFAULT_MESH_COLOR } from './3dview/helpers';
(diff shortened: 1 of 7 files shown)


packages/base/src/commands.ts
```diff
@@ -1085,27 +1085,35 @@ export function addCommands(
...
+        const content = current?.content;
+        const projection = content['Type'];
+        console.log('projection:', projection);
+        return (toggled = projection === 'Perspective');
+      } else {
+        return false;
+      }
+    },
     execute: async () => {
+      toggled = !toggled;
```
martinRenou, line 1105: We shouldn't use a global `toggled` variable here, instead we should check the current projection type in the current document.

martinRenou (changes requested):
Thanks! I have a couple of suggestions

HaudinFlorence:
@martinRenou Thanks for your comments. They have been taken into account.

SylvainCorlay (approved):

packages/base/src/widget.ts
```diff
@@ -175,7 +177,7 @@ export class JupyterCadPanel extends SplitPanel {
   }
 
   get cameraSettings(): CameraSettings | undefined {
```
martinRenou, line 179: Let's change it to `CameraSettings` everywhere and be sure it's defined?

```suggestion
  get cameraSettings(): CameraSettings {
```

martinRenou: This will eventually also clean your command a bit, removing some checks

HaudinFlorence: Ok thanks. And the same changed applies for the setter I guess?

martinRenou: Yep :) And TypeScript will then ask you to change also the interface an more places I suppose. It's good cleanup 👍🏽

packages/base/src/commands.ts
```diff
@@ -1087,25 +1057,32 @@ export function addCommands(
...
-
       if (!current) {
         return;
+      } else {
+        const currentSettings: CameraSettings = current.content.cameraSettings;
+        if (currentSettings.type === 'Perspective') {
+          current.content.cameraSettings = { type: 'Orthographic' };
+        } else {
+          current.content.cameraSettings = { type: 'Perspective' };
+        }
```
martinRenou, line 1083: Last tiny nitpick, in case we add more camera settings later, we don't want to overwrite the full settings here

```suggestion
        if (currentSettings.type === 'Perspective') {
          current.content.cameraSettings.type = 'Orthographic';
        } else {
          current.content.cameraSettings.type = 'Perspective';
        }
```

martinRenou (approved):
Thanks! 🚀

PR #719, opened 2025-04-24
 Follow up of PR 717. Fix reassignment of cameraSettings

HaudinFlorence:
Fix reassignement of cameraSettings in updateCameraSettings command. Update the labels.


martinRenou:
Works well!

Issue #718, opened 2025-04-24
Make the axes helper visible through objects like the bounding box


SylvainCorlay:
In many cases, if the objects in the scene are too big, it could appear that the axes helper button does nothing.

We should probably make the axes helper visible through objects like the bounding box.

martinRenou:
It's about passing `depthTest: false` to the material of the axes helper mesh

PR #726, opened 2025-04-29
Replace the FormDialog by a toggle button for the axes helper

HaudinFlorence:
Replace the FormDialog by a toggle button for the axes helper.
Set the material depthTest property to false in order to always render the axis on top of the shapes.

Should fix https://github.com/jupytercad/JupyterCAD/issues/718

### DIFF
--- a/packages/base/src/widget.ts
+++ b/packages/base/src/widget.ts
@@ -114,7 +114,7 @@ export class JupyterCadPanel extends SplitPanel {
   }) {
     this._view = new ObservableMap<JSONValue>({});
     const cameraSettings: CameraSettings = { type: 'Perspective' };
-    const axes: AxeHelper = { visible: false, size: 20 };
+    const axes: AxeHelper = { visible: false};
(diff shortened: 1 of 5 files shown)


packages/base/src/widget.ts
```diff
@@ -114,9 +114,11 @@ export class JupyterCadPanel extends SplitPanel {
   }) {
     this._view = new ObservableMap<JSONValue>({});
     const cameraSettings: CameraSettings = { type: 'Perspective' };
+    const axes: AxeHelper = { visible: false, size: 20 };
```
martinRenou, line 117: Let's remove the `AxeHelper.size` property from the type since we won't be using it anymore

HaudinFlorence: @martinRenou Thanks for the comment. `size` has been removed from `AxeHelper` type.

martinRenou (approved):
Thanks!!

PR #727, opened 2025-04-30
Follow up of PR 726. Fix the axes helper toggle button broken behavior

HaudinFlorence:
Fix the broken behavior for the axes helper toggle button with adding `notifyCommandChanged` in the execute method of `updateAxes` command.


martinRenou (approved):
Thanks!

PR #730, opened 2025-04-30
Update the switch mode label for the transform controls

HaudinFlorence:
Update the switch mode label for the transform controls. 
If the user is in translation mode, the label proposes to switch to the rotation mode and vice versa.


arjxn-py (approved):
Thanks :)

Issue #729, opened 2025-04-30
The console button should be a toggle


SylvainCorlay:
It already behaves like a toggle. It should probably just be styled accordingly.

This probably applies to JupyterGIS.

HaudinFlorence:
This issue has been resolved by https://github.com/jupytercad/JupyterCAD/pull/731

PR #731, opened 2025-05-02
Transform the normal console button into a toggle one

HaudinFlorence:
Transform the normal console button into a toggle one
Fixes issue https://github.com/jupytercad/JupyterCAD/issues/729


arjxn-py (approved):
Looks Good, Works Well!
Thanks @HaudinFlorence :)

PR #732, opened 2025-05-02
Apply the transform control on the parent when clicking on an edge

### LINKED ISSUE #714: Disable edge selection when the transform control button is toggled
Edge selection can make it difficult to pick an object because of edge stickiness.

In the case when the transform control button is toggled, we should priorize object selection.

I proposed disabling edge selection when the transform control button is toggled.

HaudinFlorence:
Apply the transform control on the parent when clicking on an edge
Alternative solution to disabling edge selection as suggested from https://github.com/jupytercad/JupyterCAD/issues/714, fix #714

### DIFF
--- a/packages/base/src/3dview/mainview.tsx
+++ b/packages/base/src/3dview/mainview.tsx
@@ -1308,28 +1308,46 @@ export class MainView extends React.Component<IProps, IStates> {
   private _updateTransformControls(selection: string[]) {
     if (selection.length === 1 && !this._explodedView.enabled) {
       const selectedMeshName = selection[0];
-      const matchingChild = this._meshGroup?.children.find(child =>
+
+      if (selectedMeshName.startsWith('edge')) {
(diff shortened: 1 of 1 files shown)


packages/base/src/3dview/mainview.tsx
```diff
@@ -1308,25 +1308,46 @@ export class MainView extends React.Component<IProps, IStates> {
...
+        this._transformControls.attach(selectedMesh as BasicMesh);
 
         this._transformControls.visible = this.state.transform;
         this._transformControls.enabled = this.state.transform;
 
         return;
       }
     }
-
-    // Detach TransformControls from the previous selection
```
martinRenou, line 1325: Maybe this shouldn't be removed?

HaudinFlorence: Yes. You are right. Sorry this was removed by mistake.

martinRenou (approved):
Thanks!

Issue #690, opened 2025-02-13
Notebook cell outputs don't seem to handle resize events properly


martinRenou:
https://github.com/user-attachments/assets/1d0f0e2f-a4ba-4499-924d-bdc633d4b5ee

PR #735, opened 2025-05-06
Properly handle lumino events

martinRenou:
Fix https://github.com/jupytercad/JupyterCAD/issues/690

Also fixes issues with the scene disappearing when switching tabs between notebook and cad editor

- handle lumino resize event
- handle lumino after-show event (when switching tabs)
- handle lumino after-attach event (when the widget is added)


trungleduc:
Thanks!

PR #737, opened 2025-05-14
Add 'New JupyterCAD file' option to context menu (jupytercad#634)

### LINKED ISSUE #634: "New JupyterCAD file" from the file browser
We should add a "New JupyterCAD file" from the filebrowser upon right-click, similar to the Notebook:

Image

Yahiewi:
Modified the jupytercad_core plugins.ts to add a new "New JupyterCAD File" option when opening up the file browser context menu. I gave this option a rank of 55 so it sits nicely after the "New File" and "New Notebook" options but before the "New Folder" option.

Fixes #634


martinRenou (approved):
Neat!! 

We should definitely do the same in JupyterGIS and jupytercad-freecad 🚀

PR #740, opened 2025-05-16
Add multi-edge chamfer and fillet creation (jupytercad#646)

### LINKED ISSUE #646: Can only apply chamfer/fillet one edge at a time
We should allow for applying chamfer and fillet to multiple edges at a time

Yahiewi:
Fixes #646 
As @martinRenou suggested, I modified the schemas for chamfer and fillet to make them support either multiple edges (an array of numbers) or a single edge represented by a number (for backwards compatibility with the old jcad files present in /examples for example).
Then in commands.ts, modified the getSelectedEdge() function and the operators for chamfer and fillet accordingly (by making them work for multiple edges). The new getSelectedEdges() is a bit too verbose because I had to deal with a lot of eslint issues, but feel free to change it if you'd like.
And did the same for chamfer.ts and fillet.ts.

I tested on Jupyter lab and it should be working fine. However, like I said in another issue, I can't be sure this works on the standalone version since there seem to be a lot of issues (for me at least) with it.
Also worth noting that this only works on a single object and not if you select multiple objects from different objects (even if they are identical).

### DIFF
--- a/packages/base/src/commands.ts
+++ b/packages/base/src/commands.ts
@@ -189,21 +189,28 @@ function getSelectedMeshName(
   return '';
 }
 
-function getSelectedEdge(
+function getSelectedEdges(
   selection: { [key: string]: ISelection } | undefined
-): { shape: string; edgeIndex: number } | undefined {
+): { shape: string; edgeIndices: number[] } | undefined {
   if (selection === undefined) {
     return;
   }
-
-  const selectedNames = Object.keys(selection);
-  for (const name of selectedNames) {
-    if (selection[name].type === 'edge') {
-      return {
-        shape: selection[name].parent!,
-        edgeIndex: selection[name].edgeIndex!
-      };
-    }
+  const shape = Object.values(selection)
+    .filter(
+      sel =>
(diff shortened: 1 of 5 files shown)


martinRenou:
Neat! At first glance this is exactly what I had in mind 🚀 will test it out on Monday

packages/base/src/commands.ts
```diff
@@ -189,21 +189,28 @@ function getSelectedMeshName(
...
+    .filter(
+      sel =>
+        sel.type === 'edge' && sel.parent !== undefined && sel.parent !== null
+    )
+    .map(sel => sel.parent)[0];
+  const edgeIndices = Object.values(selection)
+    .filter(
+      sel =>
+        sel.type === 'edge' &&
+        sel.parent === shape &&
```
martinRenou, line 208: Great taking care of multi-parent selection 👍🏽

martinRenou (approved):
Perfect !


https://github.com/user-attachments/assets/fdb4c4d3-d587-4768-9ee7-444606fbade0

martinRenou:
bot please update snapshots

PR #741, opened 2025-05-19
Right click drag

### LINKED ISSUE #739: "Add Annotation" shouldn't appear when holding right-click to move around the scene
Image
The "Add Annotation" option should only appear when clicking the right-mouse button and shouldn't appear when holding it to move around the scene.

Yahiewi:
Fixes #739 
This took me a while but after a few different ideas, I was able to find a really simple solution that leverages the already created event listeners for the left mouse click and the left click drag functionality. I had to extend the IMouseDrag interface by adding an optional attribute for the button that tells us whether the click was a left mouse click or a right mouse click.
The idea here was to move the command that opens the context menu out of the "contextmenu" handler and into the "mouseup" listener so that it only opens when we have clicked and released the right mouse button without dragging it more than CLICK_THRESHOLD (which is 5 pixels).

### DIFF
--- a/packages/base/src/3dview/mainview.tsx
+++ b/packages/base/src/3dview/mainview.tsx
@@ -311,7 +311,6 @@ export class MainView extends React.Component<IProps, IStates> {
       this._renderer.domElement.addEventListener('contextmenu', e => {
         e.preventDefault();
         e.stopPropagation();
-        this._contextMenu.open(e);
       });
 
       document.addEventListener('keydown', e => {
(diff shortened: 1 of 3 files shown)


packages/base/src/3dview/mainview.tsx
```diff
@@ -332,14 +331,19 @@ export class MainView extends React.Component<IProps, IStates> {
...
       this._renderer.domElement.addEventListener('mousedown', e => {
         this._mouseDrag.start.set(e.clientX, e.clientY);
+        this._mouseDrag.button = e.button;
       });
 
       this._renderer.domElement.addEventListener('mouseup', e => {
         this._mouseDrag.end.set(e.clientX, e.clientY);
         const distance = this._mouseDrag.end.distanceTo(this._mouseDrag.start);
 
         if (distance <= CLICK_THRESHOLD) {
```
martinRenou, line 341: Ah I completely forgot we solved that issue already for the object click!

martinRenou (approved):
Looks great!

PR #743, opened 2025-05-20
Fix orthographic camera near and far parameters

### LINKED ISSUE #742: Orthographic camera: Set near and far that makes sense depending on the reflength
We should set a `near` and `far` value that makes sense depending on the current reflength instead of the hardcoded values we used to have #536

Since we don't set proper near and far, the orthographic camera currently fails to render the ArchDetails file.

Empirical values may be fine.

arjxn-py:
Shall fix #742 

https://github.com/user-attachments/assets/4545931e-ae40-4bd5-9895-27c5071520e4

https://github.com/user-attachments/assets/da367daa-1ac9-4411-b298-96dba40a7530

### DIFF
--- a/packages/base/src/3dview/mainview.tsx
+++ b/packages/base/src/3dview/mainview.tsx
@@ -1078,12 +1078,6 @@ export class MainView extends React.Component<IProps, IStates> {
 
         // Position the camera at an angle away from the center
         const offset = radius * 2;
-        this._camera.position.set(
-          center.x + offset,
-          center.y + offset,
-          center.z + offset
-        );
-        this._camera.lookAt(center);
 
         // Update orthographic frustum
         this._camera.left = (-frustumSize * aspect) / 2;
@@ -1092,16 +1086,9 @@ export class MainView extends React.Component<IProps, IStates> {
         this._camera.bottom = -frustumSize / 2;
 
         // Set near/far conservatively
-        this._camera.near = 0.01;
+        this._camera.near = 0;
         this._camera.far = offset * 5;
 
-        const distance = this._camera.position.distanceTo(
-          this._controls.target
-        );
-        const zoomFactor = 1000 / distance;
-        if (updateCamera) {
-          this._camera.zoom = zoomFactor;
-        }
         this._camera.updateProjectionMatrix();
       }
 
--- a/packages/base/src/3dview/mainview.tsx
+++ b/packages/base/src/3dview/mainview.tsx
@@ -1078,6 +1078,12 @@ export class MainView extends React.Component<IProps, IStates> {
           this._camera.zoom = zoomFactor;
         }
 
+        this._camera.position.set(
+          10 * this._refLength,
+          10 * this._refLength,
+          10 * this._refLength
+        );
+
         this._camera.near = near;
         this._camera.far = far;
         this._camera.updateProjectionMatrix();
--- a/packages/base/src/3dview/mainview.tsx
+++ b/packages/base/src/3dview/mainview.tsx
@@ -1065,10 +1065,35 @@ export class MainView extends React.Component<IProps, IStates> {
         );
       }
 
-      // Update near and far for orthographic camera if applicable
       if (this._camera instanceof THREE.OrthographicCamera) {
-        const near = Math.max(this._refLength / 20, 0.01);
-        const far = this._refLength * 20;
+        const boundingSphere = new THREE.Sphere();
+        this._boundingGroup.getBoundingSphere(boundingSphere);
+        const center = boundingSphere.center;
+        const radius = boundingSphere.radius;
+
+        const aspect =
+          this._renderer.domElement.clientWidth /
+          this._renderer.domElement.clientHeight;
+        const frustumSize = radius * 2;
+
+        // Position the camera at an angle away from the center
+        const offset = radius * 2;
+        this._camera.position.set(
+          center.x + offset,
+          center.y + offset,
+          center.z + offset
+        );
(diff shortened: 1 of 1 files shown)


packages/base/src/3dview/mainview.tsx
```diff
@@ -254,28 +249,12 @@ export class MainView extends React.Component<IProps, IStates> {
...
-          this._camera = new THREE.OrthographicCamera(
-            width / -2,
-            width / 2,
-            height / 2,
-            height / -2
-          );
-          this._camera.updateProjectionMatrix();
-        }
-      }
+      this._camera = new THREE.PerspectiveCamera(
```
martinRenou, line 252: I'm not sure why we're removing the check for the current camera setting?

packages/base/src/3dview/mainview.tsx
```diff
@@ -1075,6 +1054,8 @@ export class MainView extends React.Component<IProps, IStates> {
     } else {
       this._refLength = null;
     }
+
+    this._updateCamera();
```
martinRenou, line 1058: If possible, we shouldn't reset the camera entirely here (note we're already have some code in this method that updates some camera parameters). 
Instead, we should only set the new `near` and `far` if it's an orthographic camera.

martinRenou (changes requested):

packages/base/src/3dview/mainview.tsx
```diff
@@ -1065,6 +1065,24 @@ export class MainView extends React.Component<IProps, IStates> {
...
+      // Update near and far for orthographic camera if applicable
+      if (this._camera instanceof THREE.OrthographicCamera) {
+        const near = Math.max(this._refLength / 20, 0.01);
+        const far = this._refLength * 20;
+
+        const distance = this._camera.position.distanceTo(
+          this._controls.target
+        );
+        const zoomFactor = 1000 / distance;
+        this._camera.zoom = zoomFactor;
```
martinRenou, line 1077: I'd only change the zoom level if `updateCamera` is true:

```suggestion
        if (updateCamera) {
          this._camera.zoom = zoomFactor;
        }
```

This flag was introduced to prevent camera jumping when simply editing the current scene (e.g. you wouldn't want to zoom-in zoom-out when moving objects around with the transform control).

Though I'd keep near and far setting at all times.

arjxn-py: Thanks Martin, I didn't remember what `updateCamera` was for.

packages/base/src/3dview/mainview.tsx
```diff
@@ -1065,6 +1065,26 @@ export class MainView extends React.Component<IProps, IStates> {
...
+
+        const distance = this._camera.position.distanceTo(
+          this._controls.target
+        );
+        const zoomFactor = 1000 / distance;
+        if (updateCamera) {
+          this._camera.zoom = zoomFactor;
+        }
+
+        console.log(near, far);
```
arjxn-py, line 1081:
```suggestion
```

packages/base/src/3dview/mainview.tsx
```diff
@@ -1065,6 +1065,46 @@ export class MainView extends React.Component<IProps, IStates> {
...
+        this._camera.lookAt(center);
+
+        // Update orthographic frustum
+        this._camera.left = (-frustumSize * aspect) / 2;
+        this._camera.right = (frustumSize * aspect) / 2;
+        this._camera.top = frustumSize / 2;
+        this._camera.bottom = -frustumSize / 2;
+
+        // Set near/far conservatively
+        this._camera.near = 0.01;
```
martinRenou, line 1095: Since 0 is a valid value for the orthographic camera

```suggestion
        this._camera.near = 0;
```

packages/base/src/3dview/mainview.tsx
```diff
@@ -1065,6 +1065,46 @@ export class MainView extends React.Component<IProps, IStates> {
...
+        const frustumSize = radius * 2;
+
+        // Position the camera at an angle away from the center
+        const offset = radius * 2;
+        this._camera.position.set(
+          center.x + offset,
+          center.y + offset,
+          center.z + offset
+        );
+        this._camera.lookAt(center);
```
martinRenou, line 1086:
```suggestion
```

packages/base/src/3dview/mainview.tsx
```diff
@@ -1065,6 +1065,46 @@ export class MainView extends React.Component<IProps, IStates> {
...
+        this._camera.near = 0.01;
+        this._camera.far = offset * 5;
+
+        const distance = this._camera.position.distanceTo(
+          this._controls.target
+        );
+        const zoomFactor = 1000 / distance;
+        if (updateCamera) {
+          this._camera.zoom = zoomFactor;
+        }
```
martinRenou, line 1104:
```suggestion
```

packages/base/src/3dview/mainview.tsx
```diff
@@ -1065,6 +1065,33 @@ export class MainView extends React.Component<IProps, IStates> {
...
+
+        // Update orthographic frustum
+        this._camera.left = (-frustumSize * aspect) / 2;
+        this._camera.right = (frustumSize * aspect) / 2;
+        this._camera.top = frustumSize / 2;
+        this._camera.bottom = -frustumSize / 2;
+
+        // Set near/far conservatively
+        this._camera.near = 0;
+        this._camera.far = offset * 5;
```
martinRenou, line 1090:
```suggestion
        this._camera.far = frustumSize;
```

PR #745, opened 2025-05-27
Remove UsersItem component and use the one from `@jupyter/collaboration`

### LINKED ISSUE #744: Remove UsersItem component and use the one from jupyter-collaboration instead
https://github.com/jupytercad/JupyterCAD/blob/main/packages/base/src/toolbar/usertoolbaritem.tsx#L13 should be removed and replaced by the one introduced in https://github.com/jupyterlab/jupyter-collaboration/pull/379

arjxn-py:
Shall close #744

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -138,7 +138,7 @@ jobs:
           environment-name: cad
           create-args: >-
             python=3.9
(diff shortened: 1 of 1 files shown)


martinRenou:
These diffs are the best 🚀

PR #758, opened 2025-06-06
Add rename object function

chaffra:
added rename PythonJcadObject function


python/jupytercad_lab/jupytercad_lab/notebook/cad_document.py
```diff
@@ -114,6 +114,14 @@ def remove(self, name: str) -> CadDocument:
             self._objects_array.pop(index)
         return self
 
+    def rename(self, old_name: str, new_name: str) -> CadDocument:
+        if new_name == old_name:
+            return self
+        new_obj = doc.get_object(old_name)
```
arjxn-py, line 120:
```suggestion
        new_obj = self.get_object(old_name)
```

arjxn-py (commented):
Thanks @chaffra, suggestion below shall make the CI happy :)

PR #764, opened 2025-07-07
Add STL/BREP export feature.

Yahiewi:
This PR adds the option to export jcad shapes as STL/BREP files.
image
The command can be accessed by right-clicking on a shape in the objects tree view and selecting the output file (either STL or BREP for now).
For STL files you can modify the linear and angular deflections to control the mesh resolution.

The main changes that have been made to the code are the following:
-Added support for STL as a JCadSupportedFormat in interfaces.ts, mainviewmodel.ts and actions.ts.
-Modified the schema for postOperator.json so that it now has the file type for the export.
-Modified the OCC build to compile the StlAPI_Writer function.
-Created _writeStlFile function in stlIO.ts that uses the StlAPI_Writer function to generate the STL content.
-Added a case that uses this function in postOperator.ts.
-Created worker.ts under stlplugin to receive the STL content response from the OCC worker and trigger the download.
-Created commands.ts and modified objecttree.tsx to add the context menu and the dialog form.
-Modified plugins.ts to register the workers.

### DIFF
--- a/packages/occ-worker/src/actions.ts
+++ b/packages/occ-worker/src/actions.ts
@@ -8,6 +8,7 @@ import {
 } from '@jupytercad/schema';
 
 import { getShapesFactory, ObjectFile } from './occapi';
+import { _PostOperatorForSTL } from './occapi/postOperator';
 import { OccParser } from './occparser';
 import { IOperatorArg, IOperatorFuncOutput } from './types';
 
@@ -27,9 +28,14 @@ function buildModel(
     }
     const shapeFactory = getShapesFactory();
     let shapeData: IOperatorFuncOutput | undefined = undefined;
+    // Three main paths:
+
+    // 1. Regular shapes (Part::Box, Part::Cylinder, etc.)
     if (shapeFactory[shape]) {
       shapeData = shapeFactory[shape]?.(parameters as IOperatorArg, model);
-    } else if (parameters['Shape']) {
+    }
+    // 2. Shapes from files (BREP, STL files)
+    else if (parameters['Shape']) {
       // Creating occ shape from brep file.
       const type = parameters['Type'] ?? 'brep';
       shapeData = ObjectFile(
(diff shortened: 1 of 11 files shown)


packages/occ-worker/src/actions.ts
```diff
@@ -58,6 +59,13 @@ function buildModel(
           };
           break;
         }
+        case JCadWorkerSupportedFormat.STL: {
+          shapeData = shapeFactory['Post::Operator']?.(
+            parameters as IOperatorArg,
+            model
+          );
+          break;
+        }
```
martinRenou, line 68: Small nitpick, let's reuse the other case block:

Small nitpick

```suggestion
        case JCadWorkerSupportedFormat.BREP: 
        case JCadWorkerSupportedFormat.STL: {
          shapeData = shapeFactory['Post::Operator']?.(
            parameters as IOperatorArg,
            model
          );
          break;
        }
```

packages/occ-worker/src/occapi/stlIO.ts
```diff
@@ -22,3 +22,43 @@ export function _loadStlFile(content: string): OCC.TopoDS_Shape | undefined {
...
+    const success = writer.Write(shape, fakeFileName, progress);
+    if (!success) {
+      throw new Error('StlAPI_Writer failed to write the file.');
+    }
+
+    const stlContent = oc.FS.readFile(fakeFileName, {
+      encoding: 'utf8'
+    }) as string;
+
+    console.log(`Generated STL content, length: ${stlContent.length}`);
```
martinRenou, line 55:
```suggestion
```

python/jupytercad_core/src/stlplugin/commands.ts
```diff
@@ -0,0 +1,198 @@
+import { FormDialog } from '@jupytercad/base';
```
martinRenou, line 1: Let's move these new commands in the base commands?

The `stlplugin` is to allow JupyterCAD to open STL files with the JupyterCAD viewer, it's probably not the right place to put commands for general CAD files.

PR #766, opened 2025-07-14
Bring back follow mode

arjxn-py:


brichet:
> Shouldn't this be a fix in jupyter-collaboration instead?

The default users items in jupyter-collaboration are neither clickable nor selectable, so I don't know if it make sense to have these rules there.

arjxn-py:
Thanks @martinRenou @brichet, I had followed this helpful example shared by Nicolas https://github.com/brichet/jupyter-ai/blob/optional-bot/packages/jupyter-ai/src/user-icon.tsx. 

I wonder if we implement our custom iconRenderer upstream, wouldn't we be limiting upstream `defaultIconRenderer`?


I have also updated the style of the useritem to have border and pointer cursor in [101bd56](https://github.com/jupytercad/JupyterCAD/pull/766/commits/101bd563f2527cbdeea928d68b35e8c400a98d6c)

PR #767, opened 2025-07-15
Restore the users item visibility

### LINKED ISSUE #763: UsersItem toolbar not working anymore?
I don't see the list of collaborators in the JCAD toolbar anymore, e.g. it's not possible to start the follow mode.

brichet:
Remove the `position: absolute` of the `.jp-MenuBar-anonymousIcon`, which led to a parent DOM with no size, no taken into account by the toolbar spacer.

Fix #763.


arjxn-py (approved):
Thanks :)
This looks reasonable to me. Although I could not reproduce it on main too but happy if it fixes the issue for you.

PR #784, opened 2025-09-18
Fix pressing delete key in form from deleting entire object.

asmith26:
If I wanted to remove **bad_text** from the *Object Properties > Radius 1* text field:

(NOTE: I haven't written much typescript before, so very much welcome any edits/feedback - I did wonder if I needed to include the `|| e.key === 'Backspace'` bit, also I'm happy to try writing a UI test for this if helpful?)

### Before

https://github.com/user-attachments/assets/18859d80-e026-4d61-adc0-7cfabab945ec

### After

https://github.com/user-attachments/assets/49cf8725-b249-444c-b491-88ffc0d5affa

Thanks!

### DIFF
--- a/packages/base/src/panelview/formbuilder.tsx
+++ b/packages/base/src/panelview/formbuilder.tsx
@@ -152,7 +152,9 @@ export class ObjectPropertiesForm extends React.Component<IProps, IStates> {
           className="jpcad-property-panel"
           data-path={this.props.filePath ?? ''}
           // Prevent any key press from propagating to other elements
-          onKeyDown={(e: React.KeyboardEvent) => {e.stopPropagation();}}
+          onKeyDown={(e: React.KeyboardEvent) => {
+            e.stopPropagation();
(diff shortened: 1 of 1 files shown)


packages/base/src/panelview/formbuilder.tsx
```diff
@@ -151,6 +151,12 @@ export class ObjectPropertiesForm extends React.Component<IProps, IStates> {
         <div
           className="jpcad-property-panel"
           data-path={this.props.filePath ?? ''}
+          // Prevent Delete/Backspace from propagating to other elements
+          onKeyDown={(e: React.KeyboardEvent) => {
+            if (e.key === 'Delete' || e.key === 'Backspace') {
```
martinRenou, line 156: Should we go further and prevent the propagation of all keyboard events? I feel like typing in that panel shouldn't be disrupted by other keyboard shortcuts 

@trungleduc do you have an opinion?

PR #804, opened 2025-10-15
Adds support for the `extrusion` capability to the `CadDocument` class.

asmith26:
Was trying to decide whether to call this method "extrude" vs "extrusion" - please feel free to rename if preferable. All other thoughts also very welcome.

https://github.com/user-attachments/assets/abe2a66e-b435-4ef5-8c9a-f8c348afd7dc

Thanks!

### DIFF
--- a/python/jupytercad_lab/jupytercad_lab/notebook/cad_document.py
+++ b/python/jupytercad_lab/jupytercad_lab/notebook/cad_document.py
@@ -24,11 +24,13 @@ from jupytercad_core.schema import (
     IChamfer,
     IFillet,
     ITorus,
+    ISketchObject,
     Parts,
     ShapeMetadata,
     IAny,
     SCHEMA_VERSION,
 )
+from jupytercad_core.schema.interfaces import geomLineSegment, geomCircle
 
 logger = logging.getLogger(__file__)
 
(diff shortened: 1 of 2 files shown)


python/jupytercad_lab/jupytercad_lab/notebook/cad_document.py
```diff
@@ -688,6 +688,60 @@ def intersect(
         self.set_visible(shape2, False)
         return self.add_object(OBJECT_FACTORY.create_object(data, self))
 
+    def extrude(
```
arjxn-py: +1 for the name "extrude"

python/jupytercad_lab/jupytercad_lab/notebook/cad_document.py
```diff
@@ -688,6 +688,60 @@ def intersect(
...
+        self,
+        name: str = "",
+        shape: str | int = None,
+        direction: List[float] = [0, 0, 1],
+        length_fwd: float = 10,
+        length_rev: float = 0,
+        solid: bool = False,
+        color: Optional[str] = None,
```
arjxn-py: Although I see the correct color when you extrude in the video clip on this PR but still it'd be nicer to use `"#808080"` as fallback here

asmith26:
Thanks for approving @trungleduc, sorry I can't seem to merge this, not sure if you can help with this?

Many thanks again!

Issue #806, opened 2025-10-19
Possible bug when drawing points (don't show in the .jcad document)


asmith26:
https://github.com/user-attachments/assets/6f19cfd0-82ea-41e8-9a82-504d1952e345

asmith26:
Closing (see comments in https://github.com/jupytercad/JupyterCAD/pull/807)

PR #807, opened 2025-10-31
Fix bug regarding sketch points not saved/not showing on the main canvas view

asmith26:
Fixes https://github.com/jupytercad/JupyterCAD/issues/806

- Adds missing files/code regarding `GeomPoints`.
  - When adding these I had a problem with `GeomPoints` getting written to the `.jcad` file when sketching a line/circle, consequently I had to update some of the `controlPoints` code to ensure these were excluded correctly (e.g. [here](https://github.com/jupytercad/JupyterCAD/pull/807/files#diff-3f28e8a9b16fd6f32460c9a57cf1154fb4f3490d991f2a98bf50e9948ecc8c2aR150-R151) and the edits in [this file](https://github.com/jupytercad/JupyterCAD/pull/807/files#diff-2264c0767a3b96889d53da5b4201ec281aca0e4a5d4c17aeabb9a160d4d0a967)).
  - To actually visualise points on the canvas I ended up creating very small Spheres (see [here](https://github.com/jupytercad/JupyterCAD/pull/807/files#diff-d0bcc001f1a3e40009562548ee12e1f815b89118c0299636b6e64f0d4e7c1d45)) as I couldn't find anything else suitable within OpenCascade (please do let me know if there is a better method to visualise points though).
- I've left some debug/source-maps config committed ([here](https://github.com/jupytercad/JupyterCAD/pull/807/files#diff-d608db192116352e8c419ea688e83884c6fe559e3e88b38b81442f5f54c5c1ee) and [here](https://github.com/jupytercad/JupyterCAD/pull/807/files#diff-8eb49222ec00ebb71780ff1b65695ab3b2be768d2e0b2f49b5ce27dbb95751bb)) as I found this very helpful when developing (let me know if preferable to not commit these, I just thought these development configurations might be automatically ignored when building the npm packages).
- I've tried recording/adding a UI-test for this bug fix if helpful.

All thoughts very welcome, thanks!

https://github.com/user-attachments/assets/9b3efa99-12ea-43e8-8101-99fb3edaceb4

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,13 +26,13 @@ repos:
         exclude: .bumpversion.cfg
 
   # Autoformat: Python code
-  - repo: https://github.com/psf/black
-    rev: 25.1.0
+  - repo: https://github.com/psf/black-pre-commit-mirror
+    rev: 25.9.0
     hooks:
       - id: black
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.12.11
+    rev: v0.14.3
     hooks:
       - id: ruff
         args: ['--fix']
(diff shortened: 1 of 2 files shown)


trungleduc:
Can you explain the use case of having the points in the model and 3d view?

PR #810, opened 2025-11-09
Remove `POINT` from sketcher widget

asmith26:
Following on comments in https://github.com/jupytercad/JupyterCAD/pull/807 I thought it may be helpful to remove the "POINT" tab from the sketcher widget (see the video in https://github.com/jupytercad/JupyterCAD/pull/810#issuecomment-3508893992, of course no worries if not).

Thanks!

### DIFF
--- a/packages/base/src/sketcher/sketcherwidget.tsx
+++ b/packages/base/src/sketcher/sketcherwidget.tsx
@@ -121,9 +121,6 @@ export class SketcherReactWidget extends React.Component<IProps, IState> {
         }
         break;
       }
-      case 'POINT': {
-        break;
-      }
       default:
         break;
     }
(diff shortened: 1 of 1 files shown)


trungleduc:
Having the point displayed in the Sketcher is useful for referencing the position while you are drawing other elements

asmith26:
> Having the point displayed in the Sketcher is useful for referencing the position while you are drawing other elements

Thanks for the info, I think the code still does this, e.g.:
- for `LINE` https://github.com/asmith26/JupyterCAD/blob/887b180afe8535d75b224cac645873cef5e3e6e0/packages/base/src/sketcher/sketcherwidget.tsx#L174
- for `CIRCLE` https://github.com/asmith26/JupyterCAD/blob/887b180afe8535d75b224cac645873cef5e3e6e0/packages/base/src/sketcher/sketcherwidget.tsx#L193

I've also compared the behavior visually to the [JupyterLite](https://jupytercad.github.io/JupyterCAD) version, and it appears to be the same:

https://github.com/user-attachments/assets/8a666aff-d502-4545-97c4-9941ff8d7bc6

PR #869, opened 2026-06-15
Fix: sharedModel.source should be a string, not JSON

martinRenou:
Closes #868 

Even if our file format is indeed JSON, the jupyter-server contents manager only allow text or base64. So we must keep the text format.

The only issue is that getSource in the front-end was wrongly returning a JSON structure instead of string. Leading to some issues in jupyter-collaboration.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "eslint-plugin-prettier": "5.0.1",
     "eslint-plugin-react-hooks": "^4.6.2",
     "lerna": "^7.0.0",
+    "license-webpack-plugin": "^2.3.14",
     "npm-run-all": "^4.1.5",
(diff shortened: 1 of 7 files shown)


trungleduc (approved):

martinRenou:
And that's when I give up on this PR. The CI is broken I have no clue why.

On the CI (which I can't reproduce locally) it looks like an old webpack version is pulled and put under `node_modules/@jupyterlab/builder/node_modules`. 

Rant: I hate yarn which instead of failing to resolve packages versions ends up duplicating stuff without explaining. And I hate webpack because of webpack.

martinRenou:
Replaced by https://github.com/jupytercad/JupyterCAD/pull/873

PR #2759, opened 2026-09-10
Show profile pictures in the annotation panel and the suggestion tree

arjxn-py:
Port https://github.com/geojupyter/jupytergis/pull/1843

Should be merged after jupyter collaboration version update


martinRenou (approved):
Thanks!

PR #2787, opened 2026-09-17
Disable pan and zoom while following another collaborator

### LINKED ISSUE #2779: Disable panning and zooming while following a user
When you're following someone in collaborative mode, the camera should be fully theirs - right now you can still pan and zoom, which fights the followed user's camera and makes the view jump around.

We should just disable camera controls for as long as follow mode is active, and re-enable them when you stop following.

arjxn-py:
Close #2779 

https://github.com/user-attachments/assets/ce61bf47-dfb2-4ed7-81b0-7aca08ed8235


martinRenou (approved):
Thanks!

